PR #11, opened 2022-12-21
Services provided to network

### LINKED ISSUE #10: network infrastructure could be mis-understood
the last sentence of:

````
   Infrastructure network  the network infrastructure to which a stub
      router connects.  This network can be a single link, or a network
      of links.  The network may also provide some services, such as a
      DNS resolver, a DHCPv4 server, and a DHCPv6 prefix delegation
      server, for example.
````

may be a source of confusion, as the network does not provide those services... those services are provided by the CPE to the network.

mcr:
close #10
with sentence explaining that some services are found on the infrastructure network

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -146,9 +146,10 @@
 	  the destination address is on-link.</dd>
 	<dt>Infrastructure network</dt>
 	<dd>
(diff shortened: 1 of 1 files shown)


moonshiner:
Instead of "Home Router" perhapos just say "Router"?

PR #39, opened 2024-01-30
Updates to "5.2.1 Maintenance across stub router restarts"

### LINKED ISSUE #29: Some points to address from Darren Dukes' review
> 3 – section 5.2.1 discusses stub router restarts.  I assume the two stub routers are connected to the same AIL **and** to the same stub network.  I also assume the remembered prefix time includes remembering the prefix... The text doesn’t state this clearly.

Actually, the two stub routers might not be connected to the same stub network, and that should be okay. The point is to stabilize the on-link prefix on the AIL if it's provided by a stub router. If stub routers A and B are connected to the same AIL, only one will advertise an on-link prefix. If that one reboots, the other will remember that that prefix is on-link, so it will do the right thing with responses send to hosts on that prefix. The goal is to make sure that the router that rebooted _also_ considers that prefix on-link, even though it's never seen an RA. I think this section needs a bit of work; the clarifications you've requested would be part of that, but it was useful to re-think this based on your comments as well, because I think I see how to make this work in a relatively clean way.
  

> The WG comments were good, but I would add that the remembered prefix is immediately usable again if it’s a generated ULA, as mentioned in 5.2.2.  For any other stored prefix, would we not be better off to just follow the state machine, because we don’t know if we’ve changed infrastructure networks while shut down?

If we are not advertising the prefix on the link, because there's a usable prefix on the link, and we get a message from the stub network for an address on the remembered prefix, there are the following possibilities that I can think of:

1. The prefix is a DHCP prefix, meaning it's routable and doesn't belong to the stub router. In this case, the stub router could try to renew it; if that succeeds, it could assume that the prefix is valid on the AIL and forward packets to that prefix to the AIL. If that fails, it assumes the prefix is only reachable through a router, so it forwards the packet using its routing table, and hopes for the best.
2. The prefix is a prefix generated by that stub router. In this case, if the stub router is on a different AIL, it's immaterial, since the prefix isn't routable. So the stub router can assume that that prefix is on-link and forward packets for that prefix to the link; if they reach their intended destination, great. If not, nothing is lost.
3. The prefix is a stub-network-specific prefix. As an example, Thread uses the thread XPANID, which is a 48-bit number unique to a Thread network and its set of credentials, to generate a ULA prefix for the AIL. This means that if there are several stub routers connected to the same AIL and the same Thread network, they will all always advertise the same ULA on the AIL if there's no usable prefix there. This has a really nice benefit of increasing the stability of numbering on the wire in the presence of stub router reboots. This prefix isn't routable, so if the stub router gets connected to a different AIL, there's not much it can do about it. It's not actually clear what to do in this situation, since if there's another stub router still connected to the old AIL, we have ambiguous routing to that prefix on the stub network. In the case of Thread, we don't consider this a serious concern, since thread has similar locality to home networks: if you can't reach the home router, you probably can't reach the thread network you were connected to before either.

jwhui:
Fixes #29

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -466,26 +466,20 @@
 	    and will drop the packet.
 	  </t><t>
 	    To address this problem, stub routers SHOULD remember the last time a prefix was advertised across restarts. On restart,
-	    the router can immediately begin deprecating the prefix, and can stop after the prefix valid lifetime goes to zero, based
-	    on the recorded time that the last advertisement was sent.
+	    the router configures the prefix on its interface but does not advertise it in Router Advertisements. Devices that are
+	    still using that prefix will be seen as on-link to the router, and so packets will be delivered using ND on-link rather
+	    than forwarded to the default router.
 	  </t><t>
 	    When a stub router has only flash memory with limited write lifetime, it may be inappropriate to do a write to flash
-	    every time an RA beacon containing a prefix is sent. In this case, the router SHOULD record the set of prefixes that have been advertised
-	    on infrastructure and the maximum valid lifetime that was advertised. On restart, the router should assume that hosts on
-	    the infrastructure link have received advertisements for any such prefixes, and should immediately deprecate them, and
-	    continue to do so until the maximum valid lifetime has elapsed after restart.
-	  </t>
-	  <t>
-	    [WG: we could actually just not advertise the prefix, rather than deprecating it. In this case, the host should wind
-	    up preferring some other prefix for new connections anyway, because it will have a later preferred lifetime expiry.
-	    As long as we remember the route and resume forwarding for it, existing connections can continue until the prefix
-	    becomes invalid.</t>
-	  <t>
-	    Further experience with this shows that this solution doesn't work well at all. Ideally the stub routers would all use
-	    some identifier specific to the stub network to construct the prefix advertised on the AIL. At present Thread uses the
-	    Thread "extended PAN ID", which is a 64-bit quantity. Perhaps something like the SSID could be used for a WiFi stub
-	    network. I don't think there's a way to specify this generally. But it seems as if the solution proposed here probably
-	    isn't worth doing.]</t>
+	    every time an RA beacon containing a prefix is sent. In this case, the router SHOULD record the set of prefixes that
+	    have been advertised on infrastructure and the maximum valid lifetime that was advertised. On restart, the router should
+	    assume that hosts on the infrastructure link have received advertisements for any such prefixes.
+	  </t><t>
+            When possible, it is best if all stub routers serving a particular stub network use the same prefix on the AIL. For
+            example, Thread stub routers use the 40 most-significant bits of the Thread Extended PAN ID to generate the ULA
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -466,26 +466,20 @@
...
-	    some identifier specific to the stub network to construct the prefix advertised on the AIL. At present Thread uses the
-	    Thread "extended PAN ID", which is a 64-bit quantity. Perhaps something like the SSID could be used for a WiFi stub
-	    network. I don't think there's a way to specify this generally. But it seems as if the solution proposed here probably
-	    isn't worth doing.]</t>
+	    every time an RA beacon containing a prefix is sent. In this case, the router SHOULD record the set of prefixes that
+	    have been advertised on infrastructure and the maximum valid lifetime that was advertised. On restart, the router should
+	    assume that hosts on the infrastructure link have received advertisements for any such prefixes.
+	  </t><t>
+            When possible, it is best if all stub routers serving a particular stub network use the same prefix on the AIL. For
+            example, Thread stub routers use the 40 most-significant bits of the Thread Extended PAN ID to generate the ULA
```
EskoDijk, line 479: It says Thread stub routers use 40 bits for the ULA prefix, it should say something like "the Global ID of the ULA prefix" ?  There's also 16 bits of the Subnet ID - if this is chosen differently, the prefix (/64) will differ.

So it seems vital for the intended solution here that also the Subnet ID is chosen the same. Then we have 56 bits of randomness and 8 bits for the FD00::/8 item, in total a /64.

draft-ietf-snac-simple.xml
```diff
@@ -466,26 +466,20 @@
...
-	    As long as we remember the route and resume forwarding for it, existing connections can continue until the prefix
-	    becomes invalid.</t>
-	  <t>
-	    Further experience with this shows that this solution doesn't work well at all. Ideally the stub routers would all use
-	    some identifier specific to the stub network to construct the prefix advertised on the AIL. At present Thread uses the
-	    Thread "extended PAN ID", which is a 64-bit quantity. Perhaps something like the SSID could be used for a WiFi stub
-	    network. I don't think there's a way to specify this generally. But it seems as if the solution proposed here probably
-	    isn't worth doing.]</t>
+	    every time an RA beacon containing a prefix is sent. In this case, the router SHOULD record the set of prefixes that
+	    have been advertised on infrastructure and the maximum valid lifetime that was advertised. On restart, the router should
```
EskoDijk, line 475: I read these 2 sentences here a couple of times, but don't understand this. Also don't understand what is achieved by recording the set of prefixes that have been advertised on AIL and max valid lifetime (of what?).

Abhayakara (approved):

draft-ietf-snac-simple.xml
jwhui: I updated the text to better indicate 64-bit prefix.

EskoDijk (approved):
Not all aspects described here are fully clear yet, and there's a tension between this text and other PR text that says ULA Site Prefix is used to generate the AIL ULA link prefix.

But it's best to merge the new text now and use that as a basis for further review!

PR #40, opened 2024-02-16
Updates to "5.2.2 Generating a per-stub-router ULA site prefix"

### LINKED ISSUE #31: Relax the requirement on a single 'root' ULA prefix generated by stub router
Current text (-02) has:

> A stub router MUST allocate a single ULA prefix for use in providing on-link prefixes to the stub network and the infrastructure network, as needed.

However, there may be good reasons to deviate from this. Thread for example uses a generated prefix for the AIL that uses the 64 random bits of the Extended PAN ID of the Thread Network as the bits for the ULA prefix global ID and subnet ID. This has the benefit that a more stable ULA prefix is obtained on the AIL, in the presence of rebooting / changing stub routers.

Proposal here is to relax this requirement in some way by allowing a different (random) ULA prefix for the AIL.

jwhui:
Fixes #31


draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
...
+	    Some technologies used for stub networks, for example Thread or 6LoWPAN mesh networks, can produce partitioned networks,
+            where what is notionally the same stub network winds up looking like two or more discrete links. For mesh networks, such
+            partitions can form and rejoin over time as a result of either changes in radio propagation or the addition of, or
+            removal of, devices on the mesh.
 	  </t><t>
-	    The ULA prefix allocated by a stub router SHOULD be maintained across reboots, and SHOULD remain stable over
-	    time. For privacy reasons, a stub router that roams from network to network may wish to allocate a different ULA prefix
-	    each time it connects to a different infrastructure network.
+	    On stub networks that can partition, one way of detecting that a partition has occurred is to notice that the stub
+	    router that has advertised the on-link prefix for the stub network is no longer reachable. When this occurs, stub
```
EskoDijk, line 499:
```suggestion
	    router that has advertised the on-link prefix for the stub network is no longer reachable via the stub network. When this occurs, stub
```

Need to clarify the "no longer reachable" -> specifically means no longer reachable via stub network (technology). It may still be reachable via AIL in the usual case.

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
...
+	    routers that notice this loss of reachability MUST advertise a ULA link prefix derived from their ULA site prefix on the
```
EskoDijk, line 500: before we can say this, the term "ULA site prefix" needs to be introduced first.

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
...
+	    stub network.
+	  </t><t>
+	    An implication of this is that when such a partition forms, the same ULA link (?) prefix can’t be advertised on both
+	    partitions, since this will result in ambiguous routing. We address this problem by requiring that each stub router
+	    generate its own ULA site prefix. This prefix is then available for two purposes: providing addressing on the AIL, if
+	    needed, and providing addressing on the stub network, if needed.
+	  </t><t>
+	    A stub router must therefore allocate its own ULA site prefix. ULA prefixes, described in Unique Local IPv6 Unicast
+            Addresses (<xref target="RFC4193"/>), are randomly allocated prefixes.  A stub router MUST allocate a single ULA site
```
EskoDijk, line 509: Length of ULA site prefix is not defined. RFC 4193 doesn't define "ULA site prefix".

If every stub router creates its own ULA site prefix, how can we get to the case that multiple stub routers will use the same ULA /64 prefix for the AIL as was intended by PR #39 ?

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
...
+            prefix for use in providing on-link prefixes to the stub network and the infrastructure network, as needed.
+	  </t><t>
+	    The ULA prefix allocated by a stub router SHOULD be maintained across reboots, and SHOULD remain stable over time. (TBD:
```
EskoDijk, line 512: which prefix is intended here: the ULA site prefix, or the /64 assigned to the stub network as on-link, or the /64 assigned to the AIL in case needed?

draft-ietf-snac-simple.xml
Abhayakara: The problem is that RFC 4193 doesn't really make a distinction between a ULA /48 and a ULA /64 that's derived from that ULA /48. That is, there is no terminology for the two types of ULA prefix. What we mean by "site prefix" is "ULA /48," or "the prefix identified by the ULA global identifier."

I think the way to address this is to just define the term in the glossary. I'd say something like:

ULA site prefix: a ULA /48 prefix from which allocations can be made.

ULA subnet prefix: a ULA /64 derived from a particular ULA site prefix.

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
 	    isn't worth doing.]</t>
 	</section>
 	<section>
-	  <name>Generating a ULA prefix to provide addressability</name>
+	  <name>Generating a per-stub-router ULA site prefix</name>
 	  <t>
-	    In order to be able to provide addressability either on the stub network or on an adjacent infrastructure network, a stub
-	    router must allocate its own ULA prefix. ULA prefixes, described in Unique Local IPv6 Unicast Addresses
-	    (<xref target="RFC4193"/>) are randomly allocated prefixes. A stub router MUST allocate a single ULA prefix for use in
-	    providing on-link prefixes to the stub network and the infrastructure network, as needed.
```
Abhayakara, line 496: I think this text shouldn't have been deleted.

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +488,31 @@
...
+	    Some technologies used for stub networks, for example Thread or 6LoWPAN mesh networks, can produce partitioned networks,
```
Abhayakara, line 493: It feels like this belongs in section 7.

draft-ietf-snac-simple.xml
jwhui: Restored text.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Added "ULA Site Prefix" to Glossary section.

draft-ietf-snac-simple.xml
jwhui: Added "ULA Site Prefix" and "ULA Link Prefix" to Glossary section - terms that were proposed during the last SNAC meeting.

draft-ietf-snac-simple.xml
jwhui: Updated to specify "ULA Link Prefix", which is the /64.

draft-ietf-snac-simple.xml
jwhui: Moved to Section 7.

draft-ietf-snac-simple.xml
```diff
@@ -201,6 +201,12 @@
 	<dt>RA Beacon:</dt>
 	<dd>A Router Advertisement (RA) that is multicast on a link so that hosts can see that the router is still present. This is in
 	  contrast to a unicast RA sent in response to the router solicit.</dd>
+	<dt>ULA Site Prefix:</dt>
+	<dd>A Unique Local Address /48 prefix <xref target="RFC4193"/> randomly generated by each stub router for use in providing
+	  ULA Link Prefixes to the stub network and the infrastructure network, as needed.</dd>
+	<dt>ULA Link Prefix:</dt>
+	<dd>A Unique Local Address /64 prefix derived from the ULA Site Prefix that stub routers advertise to provide addressability
+	  to the stub network and the infrastructure network, as needed.</dd>
```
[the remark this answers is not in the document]
Abhayakara: I would suggest just to make it crystal clear changing "derived from" to "allocated from". And maybe say "a total of 65,536 ULA link prefixes can be allocated from the ULA Site prefix."

draft-ietf-snac-simple.xml
```diff
@@ -675,12 +683,12 @@
...
-	for all stub routers to provide it. In the latter case, this should not be a problem: since each stub router is using
-	its own ULA prefix to provide NAT64, any 5-tuple that goes through a stub router's NAT64 translator will necessarily
-	have as its destination an IPv6 address in a particular NAT64 prefix, and that address will select the correct stub
-	router through which to send the packet for translation.</t>
+	An additional complication is that there may be more than one stub router connecting the stub network to infrastructure.  In
+	this case, it may be desirable to limit the number of stub routers providing NAT64 service, or it may be acceptable for all
+	stub routers to provide it. In the latter case, this should not be a problem: since each stub router is using its own ULA
+	Site Prefix to provide NAT64, any 5-tuple that goes through a stub router's NAT64 translator will necessarily have as its
+	destination an IPv6 address in a particular NAT64 prefix, and that address will select the correct stub router through which
+	to send the packet for translation.</t>
```
Abhayakara, line 691: Possibly worth mentioning that this also works on the return path because each stub router has its own IPv4 address, and the return packet will be destined for that IPv4 packet, and hence will always return through the stub router that translated it on the way out.

Abhayakara (changes requested):
This is nearly done once the two nits I've raised and the ones Esko raised are done.

draft-ietf-snac-simple.xml
```diff
@@ -395,13 +402,13 @@
 	    </ul>
 	  </section>
 	  <section anchor="state-deprecating">
-	    <name>Stub router deprecating its on-link prefix (STATE-DEPRECATING)</name>
+	    <name>Stub router deprecating its ULA Link Prefix (STATE-DEPRECATING)</name>
```
Abhayakara, line 405: Maybe we should leave this as it was—ULA Link prefix is ambiguous.

Abhayakara: Stub router deprecating the on-link prefix it is advertising

draft-ietf-snac-simple.xml
```diff
@@ -395,13 +402,13 @@
...
 	    <t>
 	      On entry to this state, the stub router has been treating the interface as an Advertising Interface as described in
 	      <xref target="RFC4861" section="6.2.2"/>, and MUST continue to do so.</t>
 	    <t>When the stub router has detected the availability of usable on-link prefix on the AIL to
-	      which the interface is attached, and that prefix is preferable to the one it is advertising, it continues to advertise
-	      its own prefix, but deprecates it:
+	      which the interface is attached, and that prefix is preferable to the ULA Link Prefix it is advertising, it continues to advertise
+	      its ULA Link Prefix, but deprecates it:
```
Abhayakara, line 411: stet

draft-ietf-snac-simple.xml
```diff
@@ -513,12 +521,12 @@
 	    routing fabric, so even though packets might be able to be forwarded to the intended destination, there would be no
 	    return path. So when the only prefix that is available is the one provided by the stub router, cloud services will
 	    not be reachable via IPv6, and infrastructure-provided NAT64 will not work. Therefore, when the stub router is able to
-	    successfully acquire a prefix using DHCPv6 PD, it MUST use DHCPv6 PD rather than its own self-generated ULA prefix.
+	    successfully acquire a prefix using DHCPv6 PD, it MUST use DHCPv6 PD rather than its own self-generated ULA Site Prefix.
```
Abhayakara, line 524: We could say here "rather than the ULA Link prefix it allocated for the stub network out of its ULA Site Prefix."

draft-ietf-snac-simple.xml
```diff
@@ -201,6 +201,12 @@
 	<dt>RA Beacon:</dt>
 	<dd>A Router Advertisement (RA) that is multicast on a link so that hosts can see that the router is still present. This is in
 	  contrast to a unicast RA sent in response to the router solicit.</dd>
+	<dt>ULA Site Prefix:</dt>
+	<dd>A Unique Local Address /48 prefix <xref target="RFC4193"/> randomly generated by each stub router for use in providing
+	  ULA Link Prefixes to the stub network and the infrastructure network, as needed.</dd>
```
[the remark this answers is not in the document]
jwhui: Updated.

draft-ietf-snac-simple.xml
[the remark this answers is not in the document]
jwhui: Updated

draft-ietf-snac-simple.xml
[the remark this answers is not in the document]
jwhui: Updated.

draft-ietf-snac-simple.xml
```diff
@@ -488,16 +495,17 @@
...
 	<section>
-	  <name>Generating a ULA prefix to provide addressability</name>
+	  <name>Generating a per-stub-router ULA Site Prefix</name>
 	  <t>
 	    In order to be able to provide addressability either on the stub network or on an adjacent infrastructure network, a stub
-	    router must allocate its own ULA prefix. ULA prefixes, described in Unique Local IPv6 Unicast Addresses
-	    (<xref target="RFC4193"/>) are randomly allocated prefixes. A stub router MUST allocate a single ULA prefix for use in
+	    router MUST allocate its own ULA Site Pefix. ULA prefixes, described in Unique Local IPv6 Unicast Addresses
+	    (<xref target="RFC4193"/>) are randomly allocated prefixes. A stub router MUST allocate a single ULA Site Prefix for use in
 	    providing on-link prefixes to the stub network and the infrastructure network, as needed.
```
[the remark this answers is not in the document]
jwhui: Updated.

draft-ietf-snac-simple.xml
```diff
@@ -513,12 +521,12 @@
...
 	    return path. So when the only prefix that is available is the one provided by the stub router, cloud services will
 	    not be reachable via IPv6, and infrastructure-provided NAT64 will not work. Therefore, when the stub router is able to
-	    successfully acquire a prefix using DHCPv6 PD, it MUST use DHCPv6 PD rather than its own self-generated ULA prefix.
+	    successfully acquire a prefix using DHCPv6 PD, it MUST use DHCPv6 PD rather than its own self-generated ULA Site Prefix.
           </t><t>
             A stub router SHOULD request stub network prefixes with length 64. If the stub router obtains a prefix with length less
             than 64, it SHOULD generate a /64 from the obtained prefix by padding with zeros. If the stub router obtains a prefix
-            with length greater than 64, the stub router MUST treat the prefix as unusable and allocate a prefix out of its ULA
-            prefix instead.
+            with length greater than 64, the stub router MUST treat the prefix as unusable and allocate a prefix out of its ULA Site
```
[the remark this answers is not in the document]
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
```diff
@@ -206,6 +206,13 @@
 	<dt>RA Beacon:</dt>
 	<dd>A Router Advertisement (RA) that is multicast on a link so that hosts can see that the router is still present. This is in
 	  contrast to a unicast RA sent in response to the router solicit.</dd>
+	<dt>ULA Site Prefix:</dt>
+	<dd>A Unique Local Address /48 prefix <xref target="RFC4193"/> randomly generated by each stub router for use in providing
+	  ULA Link Prefixes to the stub network and the adjacent infrastructure link, as needed.</dd>
+	<dt>ULA Link Prefix:</dt>
+	<dd>A Unique Local Address /64 prefix allocated from the ULA Site Prefix that stub routers advertise to provide
```
Abhayakara, line 213: this is now a bit hard to parse. How about:
allocated from the ULA site prefix. Stub routers can use ULA Link prefixes to provide addressability on the stub network and/or adjacent infrastructure link as needed. If a BR is doing NAT64, the NAT64 prefix is also a ULA Link Prefix. A total of....

draft-ietf-snac-simple.xml
```diff
@@ -206,6 +206,13 @@
 	<dt>RA Beacon:</dt>
 	<dd>A Router Advertisement (RA) that is multicast on a link so that hosts can see that the router is still present. This is in
 	  contrast to a unicast RA sent in response to the router solicit.</dd>
+	<dt>ULA Site Prefix:</dt>
+	<dd>A Unique Local Address /48 prefix <xref target="RFC4193"/> randomly generated by each stub router for use in providing
+	  ULA Link Prefixes to the stub network and the adjacent infrastructure link, as needed.</dd>
```
Abhayakara, line 211: providing should be allocating. "as needed" is probably not making this clearer.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
```diff
@@ -692,12 +703,14 @@
...
       <t>
-	An additional complication is that there may be more than one stub router connecting the stub network to infrastructure.
-	In this case, it may be desirable to limit the number of stub routers providing NAT64 service, or it may be acceptable
-	for all stub routers to provide it. In the latter case, this should not be a problem: since each stub router is using
-	its own ULA prefix to provide NAT64, any 5-tuple that goes through a stub router's NAT64 translator will necessarily
-	have as its destination an IPv6 address in a particular NAT64 prefix, and that address will select the correct stub
-	router through which to send the packet for translation.</t>
+	An additional complication is that there may be more than one stub router connecting the stub network to infrastructure.  In
+	this case, it may be desirable to limit the number of stub routers providing NAT64 service, or it may be acceptable for all
+	stub routers to provide it. In the latter case, this should not be a problem: since each stub router is using its own ULA
```
Abhayakara, line 708: Maybe a paragraph break before "In the latter case"?

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Updated.

draft-ietf-snac-simple.xml
jwhui: Updated.

Abhayakara (approved):

PR #44, opened 2024-02-26
Correct locations of flags in RA/PIO and add M/O flag copy requirement

### LINKED ISSUE #32: Should a stub router learn RA header parameters from other routers?
Say that here's a WiFi network. 

- The WiFi AP is a DHCPv6 server advertising RAs with 'M' flag set to 1 in the header. 
- There's a Thread Border Router (a stub router) in the WiFi network using the WiFi as the AIL.

**Question**: Should the stub router learn the RA headers from the WiFi AP and set the 'M' flag to 1 in its own RA header? If so, are there any other header parameters that it should learn from the WiFi AP's RA?

I wonder if it will cause any problems if the stub router just send RAs with 'M' and 'O' flags set to 0. In [RFC4861](https://www.rfc-editor.org/rfc/rfc4861.html#section-4.2) it says:
```
        Note: If neither M nor O flags are set, this indicates that no
        information is available via DHCPv6.
```
This makes the RA (M=0, O=0) sent by the stub router contradicts with the RA (M=1,O=0) sent by WiFi AP. 

In [RFC4861](https://www.rfc-editor.org/rfc/rfc4861.html#section-6.3.4) it talked about how a host should handle RAs from different sources. 
```
   When multiple routers are present, the information advertised
   collectively by all routers may be a superset of the information
   contained in a single Router Advertisement.  Moreover, information
   may also be obtained through other dynamic means like DHCPv6.  Hosts
   accept the union of all received information; the receipt of a Router
   Advertisement MUST NOT invalidate all information received in a
   previous advertisement or from another source.  However, when
   received information for a specific parameter (e.g., Link MTU) or
   option (e.g., Lifetime on a specific Prefix) differs from information
   received earlier, and the parameter/option can only have one value,
   the most recently received information is considered authoritative.
```

However, it's not very clear to me how the host should handle the M flag from different sources:
1. Use the union of the M flags. That probably means the host thinks there's a DHCPv6 server on the Wi-FI network. 
or
2. Respect the M flag from the most recently received RA header.

EskoDijk:
* close #32 
* also corrected ULA -> ULA prefix ('A' stands for address so can't be a prefix itself)


Abhayakara (approved):

draft-ietf-snac-simple.xml
```diff
@@ -354,9 +354,14 @@
...
 	      In this state, the stub router generates its own on-link prefix for the interface. This prefix has a valid and
 	      preferred lifetime of STUB_PROVIDED_PREFIX_LIFETIME seconds. The stub router sends a router advertisement containing
-	      this prefix. The 'A' (autonomous configuration), 'L' (on-link) <xref target="RFC4861" section="4.6.2"/> and the Stub
-              Router bit (<xref target="I-D.hui-stub-router-ra-flag"/>) MUST be set in the Router Advertisement header flags
-              field <xref target="RFC5175"/>.</t>
+	      this prefix in a Prefix Information Option (PIO). In the PIO, the 'A' (autonomous configuration) and 'L' (on-link) 
+	      flags <xref target="RFC4861" section="4.6.2"/> MUST be set. The Stub
+	      Router bit (<xref target="I-D.hui-stub-router-ra-flag"/>) MUST be set in the Router Advertisement header flags
+	      field <xref target="RFC5175"/>. In the Router Advertisement header, the values of the 'M' and 'O' flags MUST be copied
+	      from the respective M/O values seen in a recent Router Advertisement received from a non-stub-router, if any.
```
Abhayakara, line 361: "seen in the most recent"

Abhayakara: Maybe we should add "After a reboot, if no recent advertisements have been received from non-stub routers, but a recent advertisement has been received from a stub router, the values for the 'M' and 'O' bits provided by that stub router should be copied. After STALE_RA_TIME after reboot, the stub router should go back to the behavior of only using 'M' and 'O' bit values from non-stub routers. This avoids a situation where a stub router that has rebooted starts to advertise different 'M' and 'O' bits than other stub routers present on the link."

draft-ietf-snac-simple.xml
```diff
@@ -354,9 +354,14 @@
 	    <t>
 	      In this state, the stub router generates its own on-link prefix for the interface. This prefix has a valid and
 	      preferred lifetime of STUB_PROVIDED_PREFIX_LIFETIME seconds. The stub router sends a router advertisement containing
-	      this prefix. The 'A' (autonomous configuration), 'L' (on-link) <xref target="RFC4861" section="4.6.2"/> and the Stub
-              Router bit (<xref target="I-D.hui-stub-router-ra-flag"/>) MUST be set in the Router Advertisement header flags
-              field <xref target="RFC5175"/>.</t>
+	      this prefix in a Prefix Information Option (PIO). In the PIO, the 'A' (autonomous configuration) and 'L' (on-link) 
+	      flags <xref target="RFC4861" section="4.6.2"/> MUST be set. The Stub
+	      Router bit (<xref target="I-D.hui-stub-router-ra-flag"/>) MUST be set in the Router Advertisement header flags
```
Abhayakara, line 359: The registry name is "IPv6 ND Router Advertisement flags"

Abhayakara: Worth clarifying that the the 'L' bit may not be appropriate for some layer-2 implementations, the exact handling of such situations is out of scope for this document. In such cases setting the 'L' bit might not be appropriate.

EskoDijk:
We discussed in the call that STALE_RA_TIME = 10 minutes may be too low as a limit for considering the most recent RA with M/O flags as "not applicable anymore".  An alternative could be to use the "Router Lifetime" in the RA header, which indicates the lifetime of the router in its role as a default-router. This has a maximum value at 18.2 hours. However, there can be cases where a router is sending RA (with M and O flag values) *and* the Router Lifetime has a value 0 (indicating it's not a default router). So this solution can't work in general.

Another solution that was discussed is using the valid or preferred lifetime from the PIO option, which has a disadvantage that it can be set very large (e.g. some implementations may use 1 month, or even infinity) and also the disadvantage that a PIO may not even be present in the RA which has the M/O flags set.

Something like 30 minutes may be appropriate e.g. 3 * STALE_RA_TIME.
This fits roughly with the proposals made in https://datatracker.ietf.org/doc/html/draft-ietf-6man-slaac-renum#section-4 : 
it defines ND_DEFAULT_PREFERRED_LIFETIME which can be 30 minutes (using defaults from RFC 4861), or 45-90 minutes (using BCP202 ) to put a "cap" on overly long advertised lifetimes.

Abhayakara:
BTW, the document I was referring to is this one: https://www.ietf.org/archive/id/draft-ietf-6man-slaac-renum-07.html

It looks like it's expired—not sure what happened.

jwhui:
RFC 4861 specifies:
```
      MaxRtrAdvInterval
                     The maximum time allowed between sending
                     unsolicited multicast Router Advertisements from
                     the interface, in seconds.  MUST be no less than 4
                     seconds and no greater than 1800 seconds.
```
So I don't think it can be less than 1800 seconds for timing out the received M/O state. Unless we consider sending a Router Solicitation to trigger RA transmissions.

Abhayakara:
Oh, that's a good observation, Jonathan. Let's use that number. :)

(That's the number I wanted anyway—I just didn't have a good justification for it!)

draft-ietf-snac-simple.xml
EskoDijk: Text for "after reboot" was included and reworded.

draft-ietf-snac-simple.xml
EskoDijk: L flag exception case was added; and "IPv6 ND Router Advertisement flags" was shortened to "Router Advertisement flags" because the existing terminology in the draft is not to say "IPv6 ND Router Advertisement" all the time, but just "router advertisement" typically, or "RA".

EskoDijk:
PR now updated and rebased. Because 18.2 hours due to RFC 8319 may be a little long, I've added a new constant MAX_FLAGS_COPY_TIME for this set to 150 minutes. (We may discuss and shorten to 60, 50 or 30 mins also)

Also I've added another max cap on the "copy time", which is given by the Router Lifetime that the non-stub-router advertises, if it's non-zero.   Reason for this is that MaxRtrAdvInterval is by RFC 4861/8319 always lower than the set Router Lifetime value (if >0).  So the Router Lifetime value provides a max cap on MaxRtrAdvInterval.  Given this information, it would be incorrect I think not to use it.  This can still be up to 18.2 hours, which is very long, so then the MAX_FLAGS_COPY_TIME is then activated to stop the copying earlier.    In "typical router" cases the Router Lifetime would be set to at most 9000 s (2.5 hours) following the original unamended RFC 4861, perhaps?  The 150 minutes == 9000 s of MAX_FLAGS_COPY_TIME now coincides with this.

Abhayakara:
I'm actually okay with using 18.2 hours as the baseline. I think it's a bad idea to use this interval as the router advertisement interval, but it's not really our problem. If the network is configured this way, I think it's okay to use that number.

EskoDijk:
> Unless we consider sending a Router Solicitation to trigger RA transmissions.

@jwhui This is also a good point to consider: in current wording there's no requirement to send RS for this purpose, but a stub router could do it still (if nothing comes back then the status quo is maintained).  Current text also captures cases of any unicast (/multicast) RAs are processed that the stub router receives as a result of it sending RSs for another reason (e.g. maybe it's trying to route packets from the stub to a next hop). So I don't see a need to say more here about sending RS only for this purpose.

jwhui (approved):

PR #46, opened 2024-02-26
State Type C host assumption

### LINKED ISSUE #35: Clarify SNAC solution assumes the IPv6 host on AIL is a Type C Host
Based on comment: https://mailarchive.ietf.org/arch/msg/snac/SXIt8RGKYyqDsg41Kucg-qVhIj4/
and responses: https://mailarchive.ietf.org/arch/msg/snac/ZXCDdBUjFn39Vr6sc5lZiEqo4NE/ https://mailarchive.ietf.org/arch/msg/snac/IDvew688_weuF_ePVxfpEJskxv4/

it is clear that the SNAC stub router solution may in some cases not work if the IPv6 Host on the AIL that needs to send an IPv6 packet to a host on a stub network is a Type A or Type B Host.  In fact, our assumption so far has been that it is a Type C host.

Proposal: add clarification that solution scope is intended for Type C hosts on the AIL.  (Though it might work for Type A and B hosts in particular cases, this is not in our scope.)
Refer to RFC 4191 Section for the 3 host types.

(FYI @auerswal)

jwhui:
Fixes #35

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -151,6 +151,11 @@
 	  Reachability from hosts on adjacent infrastructure links means that when host (A) on an adjacent infrastructure link has a datagram destined for the IP
 	  address of a host (B) on the stub network, a next-hop router is known by host (A) such that, when the datagram is sent to
 	  that router, it will ultimately reach host (B) on the stub network.</t>
+	<t>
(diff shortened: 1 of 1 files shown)


EskoDijk (approved):
Looks like a good place to add this scoping information - in the goals section. Thanks

Abhayakara (approved):

Abhayakara (approved):

PR #53, opened 2024-03-19
Add an appendix to address issue 52, the topology taxonomy

### LINKED ISSUE #51: Relax the requirement "'L' bit is set" for particular cases
In -04, "'L' bit is set" is required for a suitable prefix. However section 5.1.2.3 allows sending PIO with L flag cleared, in particular cases.

This leads to the issue that the generated PIO AIL prefix won't be recognized as a "suitable prefix" by other stub routers. So the requirement "'L' bit is set" must be relaxed for particular link layer technologies where L=0.

Abhayakara:
Fixes #51
Fixes #56

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -306,7 +306,12 @@
 	    A prefix is not considered a suitable on-link prefix if the 'L' bit is set, but the 'A' bit is not set. This indicates
 	    that node addressability is being managed using DHCPv6. Nodes are not required to use DHCPv6 to acquire addresses, so a
 	    prefix that requires the use of DHCPv6 can't be considered "suitable"&mdash;not all hosts can actually use it.
-	  </t><t>
+	  </t>
+	  <t>
+	    Note: there can be layer two networks where neighbor discovery is not supported and therefore we cannot set the 'L' bit,
+	    but could set the 'A' bit. The behavior of stub networks when connecting to such networks is out of scope for this document.
+	  </t>
+	  <t>
 	    A prefix is considered to be advertised on the link if, when a Router Solicit message
 	    (<xref target="RFC4861" section="4.1" sectionFormat="comma"/>) is sent, a Router Advertisement message is received in
 	    response which contains a prefix information option (<xref target="RFC4861" section="4.6.2" sectionFormat="comma"/>)
@@ -871,6 +876,33 @@
 	  This document describes how to implement a stub router such that it operates correctly in this situation, whether the ISP is providing IPv6
 	  reachability to the Internet or not.
 	</t>
+	<t>
+	  In some unmanaged network settings, there is a "guest" network in addition to the main network. In this configuration, if
+	  a stub router is added to the guest infrastructure network, no communication will be possible. The general intended
+	  behavior of the guest network is to isolate untrusted hosts. Since this would be the intended behavior on the part of the
+	  owner of the network, it won't be a surprise to them, since they had to explicitly give the stub router's owner the guest
+	  network credentials and not the main network credentials. This should also mean that the owner of the stub router will not
+	  expect it to function in this scenario.
+	</t>
+	<t>
+	  An additional feature of some unmanaged networks is that the owner of the network can choose to isolate all devices on the
+	  network, so that devices on the network are able to use the Internet, but not communicate with each other. This is essentially
+	  the same as being connected to the guest network, except that there is no other network. In this case we would expect that the
+	  owner of the network doesn't expect any devices attached to the network to be able to communicate with any other device, so
+	  the failure of devices connected to infrastructure to communicate with devices on the stub network would not be a surprise.
+	  The owner of the stub router might be surprised in this case, but ultimately the owner of the infrastructure network gets
+	  to make this decision, and there isn't anything we can or should do on behalf of the stub router's owner in this case.
+	</t>
+	<section>
+	  <name>Use on an unmanaged (non-home) IPv6 network</name>
+	  <t>
+	    In this scenario we have a site that is not a home, so perhaps a restaurant or business, where there is no network
+	    operator per se, and the network is deployed similarly to a home network. There is little difference between this
+	    scenario and an unmanaged home network, but expectations may be different. In particular, it is very common in such
+	    settings for there to be a guest network for visitors, or for the network to enforce isolation between all nodes
+	    connected to it.
+	  </t>
+	</section>
       </section>
       <section>
 	<name>Use on a managed network</name>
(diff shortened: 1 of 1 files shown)


EskoDijk:
@Abhayakara The new sections need to go in just after all the "<references>", otherwise the XML doesn't build (order is fixed).

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+	  PD-per-host (where hoses are encouraged to use prefix delegation) or just ordinary prefix delegation (where hosts are given prefix
+	  delegation if they ask for it, but not encouraged to ask for it).
+	</t>
+	<t>
+	  In such a situation, if mDNS works on the infrastructure link, the stub router will function correctly, because the delegated prefix
+	  will be correctly routed.
+	</t>
+	<t>
+	  It's worth noting that IPv4 devices that act similarly to stub routers, using NAT64, already exist and may indeed use the stub network
+	  functionality to support internal connections that aren't even apparent to the user. So it's not implausible that it might be desirable
```
EskoDijk, line 890: Sentence is very complex "not implausible that it might be desirable to enable ... "

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+	  to enable such a use case for IPv6, and the current stub network document does in fact enable this use case.
```
EskoDijk, line 891:
```suggestion
	  to enable such a use case for IPv6, and the present document does in fact enable this use case.
```

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+	  the stub router will advertise an on-link prefix that permits autonomous configuration. Because the stub router is not advertising a
+	  default route, this should not cause any operational issues.
+	</t>
+	<t>
+	  It might be surprising to the network operator to see the stub router's prefix advertised on the link. However, a network
+	  operated by a knowledgable operator that does not implement RA guard can always find itself in this situation. This does
+	  not create any actual operational problem.
+	</t>
+	<t>
+	  We need to address the case where the network operator is not knowledgable and did not knowingly configure a unsuitable
```
EskoDijk, line 915: complex formulation: "did not knowingly configure a unsuitable prefix, ... "   - could this be simplified or split?

Was the operator not aware of the existence of SNAC? There can also be knowledgeable operators that are not aware that SNAC exists. Does it matter for the behavior here whether the operator knows a lot or knows a little? It's how the network was configured, that counts.

EskoDijk: The caption says "unmanaged" so why is there a network operator? (There isn't, it that case, right?)
Or is "badly managed" intended?

EskoDijk:
Question on the cases included: isn't there a case possible where

* network is managed and is IPv6
* only DHCPv6 addressing is used (as intended by operator) - assume DHCPv6 hands out ULA addresses
* no RA-Guard is applied
* user brings stub router, plugs it in, and it advertises a ULA prefix in PIO
* other devices on the link start configuring a SLAAC IPv6 address and try to use this to communicate (and routing fails because the DHCP managed ULA was not used as the source, but rather the SLAAC based ULA)

But maybe this is not relevant because the source address selection algorithm would always pick the correct, managed ULA due to a better prefix match with the IPv6 destination address ?

draft-ietf-snac-simple.xml
Abhayakara: A network always has an operator. The operator may or may not do any actual "management." In a typical home network, the "operator" is the home user. The home user is not expected to do any "management" of the network. Possibly this could be made clearer. :)

Abhayakara:
Regarding the DHCPv6-only use case, we are assuming that the network operator wants to control which devices have addresses on the local link. In this case, they MUST use RA Guard. If they do not, then there is no protection against random devices sending RAs. So this would be a misconfiguration, and hence we don't need to do anything about it.

EskoDijk:
@Abhayakara Right, I also agree we don't have to do anything about solving this "problem" other than pointing to RA guard.  But still I think it's worth mentioning: it's the one identified corner case where the SNAC router has unwanted side effects.  I wasn't even sure if I was correct in predicting the effects.

Considering that there is I think no other IETF-specified / IETF-compliant device that, when plugged in, would instantly disrupt this network in the way that the stub router does. 

If this is the worst case we can find, it should actually show that our stub router is good to go.

draft-ietf-snac-simple.xml
EskoDijk: Hmm, yes good to clarify this. The person who last plugged the Ethernet cable in or our may be the "operator" then. Such a typical operator isn't the type of person that would be surprised to see a particular prefix on the link because they don't look at tools showing prefixes and they don't know what these things are.   So that gave me a different impression of what an "operator" is here.

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+	<t>
+	  In this scenario, a non-expert home user connects stub router to own unmanaged home network. This is the key intended use case for stub networks.
+	  This document describes how to implement a stub router such that it operates correctly in this situation, whether the ISP is providing IPv6
+	  reachability to the Internet or not.
+	</t>
+      </section>
+      <section>
+	<name>Use on properly managed network</name>
+	<t>
+	  In this scenario, a non-expert user attaches a stub router to an infrastructure network that's managed. This network has correctly deployed
```
jwhui, line 875: nit: is "correctly" useful here?
```suggestion
	  In this scenario, a non-expert user attaches a stub router to an infrastructure network that's managed. This network has deployed
```

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+	  will be correctly routed.
+	</t>
+	<t>
+	  It's worth noting that IPv4 devices that act similarly to stub routers, using NAT64, already exist and may indeed use the stub network
+	  functionality to support internal connections that aren't even apparent to the user. So it's not implausible that it might be desirable
+	  to enable such a use case for IPv6, and the current stub network document does in fact enable this use case.
+	</t>
+      </section>
+      <section>
+	<name>Use on a managed network without IPv6</name>
```
jwhui, line 895: Is it clear to readers what "network with IPv6" and "network without IPv6" means?

draft-ietf-snac-simple.xml
```diff
@@ -859,6 +859,66 @@
...
+      <section>
+	<name>Unmanaged home network</name>
+	<t>
+	  In this scenario, a non-expert home user connects stub router to own unmanaged home network. This is the key intended use case for stub networks.
+	  This document describes how to implement a stub router such that it operates correctly in this situation, whether the ISP is providing IPv6
+	  reachability to the Internet or not.
+	</t>
+      </section>
+      <section>
+	<name>Use on properly managed network</name>
```
darrendukes, line 873:
```suggestion
	<name>Use on managed network</name>
```

darrendukes (approved):

PR #56, opened 2024-05-09
Change Glossary to Terminology and add more terms

### LINKED ISSUE #47: Should we add a reference to RFC 8200 in the Glossary section?
* we use various terms defined there like "node", "router", "host"
* however, making this reference explicit is probably uncommon in IETF drafts (readers are supposed to know this already) - just wanted to check with the authors.

darrendukes:
Close #47 
Added Router, Node Host and change Glossary to Terminology.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -166,8 +166,14 @@
     </section>
 
     <section>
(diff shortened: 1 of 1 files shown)


Abhayakara:
Fixes #51
Fixes #56

PR #63, opened 2024-05-23
Fix build by moving references sections up and adding RFC 8200 reference.

EskoDijk:
Note: xml2rfc requires <references> sections to be before <section> elements in backmatter.
With this, `make` works again.


EskoDijk:
Will be merged shortly (after the SNAC interim call)!

PR #66, opened 2024-05-23
Add mandatory sections: IANA, Security. Close #50

### LINKED ISSUE #50: Add Security Considerations and IANA Considerations sections
For security we may add:
* risk of advertising self-generated ULA prefix? Any specifics?
* if site employs technology like RA-guard, the RAs of our unmanaged stub router will be blocked

EskoDijk:
In internet drafts, the IANA Considerations and Security Considerations sections are mandatory.
Exact text for security can be adapted also later on if needed.

close #50

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -880,14 +880,9 @@
     <section>
       <name>Security Considerations</name>
       <t>
-        By design, a stub router starts operating as an IPv6 router if no existing (suitable) IPv6 configuration is found on the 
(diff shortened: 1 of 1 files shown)


EskoDijk:
@Abhayakara I've removed the RA Guard text.  It was there not because SNAC-routers introduce a security issue, but because we think that there's a group of people/network-administrators/IETFers/others who consider any device that can autostart acting as IPv6 router and can autostart sending ND message on a link as a "security hazard" even if it's not.
Mentioning RA Guard there could be a proactive way to hopefully avoid too many discussions around this during future document reviews.

PR #79, opened 2024-06-20
RA content made explicit

darrendukes:
Closing issue#36
This PR is not building, so need to clean it up before commit.
Comments appreciated during that process.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -871,6 +871,20 @@
 	  can communicate with IPv4 hosts on the infrastructure network and the global internet.</dd>
       </dl>
     </section>
+    <section>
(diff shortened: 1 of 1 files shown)


EskoDijk:
@darrendukes PR #63 is still needed before anything can build ;-)

EskoDijk:
@darrendukes @Abhayakara  sorry I didn't intend to close this PR, Github did that somehow after me pushing to the branch. But it's good to have this one closed and replaced by #81  , because the merge direction was specified here the wrong way around.

The build is now fixed in the branch `36-make-all-contents......` and also updated to make it merge-able into main.

I prefer to merge rather sooner than later - nits can be corrected with future PRs easily.

PR #81, opened 2024-06-21
RA content made explicit (FIXED PR!)

### LINKED ISSUE #36: Make all contents of RA sent by stub router explicit, and with rationale
Based on email discussion: https://mailarchive.ietf.org/arch/msg/snac/cST1bdMJ7g52nnws0Si71jAF7io/
(lower part)

It's proposed to make all the elements of the stub router RA explicit, including rationale for the settings.

This would also include the M / O flag settings of #32 

(FYI @auerswal)

EskoDijk:
This is a FIXED version of the PR https://github.com/ietf-wg-snac/draft-ietf-snac-simple/pull/79 , which appears to be created the wrong way around! (not merging into main, but merging *from* main).

We should close/drop the other PR I think to avoid a Git mess.

FYI @darrendukes @Abhayakara 

Close #36

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -169,7 +169,7 @@
     </section>
 
     <section>
-      <name>Terminology</name>
+      <name>Glossary</name>
       <dl>
 	<dt>Node:</dt>
 	<dd>A device that implements IPv6.</dd>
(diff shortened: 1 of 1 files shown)


EskoDijk:
With the latest commit, the build is again fixed and also rebased w.r.t. latest changes in 'main' branch.

To make the build work, some broken/non-existing references were converted to informative text with TBD labels. This can be fixed after merging, once the correct references become available.

Issue #38, opened 2023-11-01
Review if any items of draft-ietf-v6ops-dhcp-pd-per-device-04 are useful / applicable to stub router


EskoDijk:
See draft: https://datatracker.ietf.org/doc/html/draft-ietf-v6ops-dhcp-pd-per-device-04

which just passed WGLC today.

This has some potentially useful considerations and requirements related to the SNAC stub router. In particular,

* Section 8: PD prefix length - required to be /64 or smaller
* Section 9: Client Mobility - do we consider this? Also relates to #37 requirements for DHCPv6-PD rebind/renew.
* Section 11: refers to a new PIO flag that indicates DHCPv6-PD is "preferred" in this network.  Should a SNAC stub router do anything with this flag (yet)?
* Section 11: considerations when running out of PD prefixes - e.g. in a home network.  What would a SNAC stub router do if no more prefixes are available? (e.g. fallback to self-generated prefix for stub network)

EskoDijk:
Note the draft that defines the new "P" flag for PIO in RA to signal that DHCPv6-PD is preferred over SLAAC is https://datatracker.ietf.org/doc/html/draft-collink-6man-pio-pflag-03

Abhayakara:
Regarding section 11, this has no bearing on what we do for the stub network, but does bear on what we do for the AIL. If the AIL provides PD-per-host, then this means we don't need to advertise a prefix with the 'A' bit set on the AIL: PD-per-host addresses our concern, because Android will support PD-per-host and everything else supports at least NA-per-host.

Abhayakara:
Regarding section 9 as mentioned above, we have a separate issue for that, so we will not address that point in this issue.

Abhayakara:
So the last two points Esko mentions above are what needs to be addressed in this issue. We should also make sure that the first point is addressed, but I think it already is.

EskoDijk:
@Abhayakara Indeed the first point on /64 prefix length is addressed in the current -04 draft already, Section 5.2.3.

On the handling of the P flag: if the P flag is set (meaning PD-preferred-over-SLAAC) - I don't think this has any impact on the determination of a suitable prefix on the AIL, and the generation of a suitable prefix by the stub router.

Suppose the A flag = 0 and P flag = 1:  to get an address, DHCPv6-PD would be required. And we have our basic assumption in 5.1.1:

```
[...]

PR #92, opened 2024-10-07
Allow either the P-flag or A-flag to make a prefix usable

Abhayakara:
Add text to the suitable prefix section that allows a prefix that is local and has either the A flag or the P flag set to be considered usable, whereas before only prefixes with the A flag set were considered usable.


Abhayakara:
Closes #38

Issue #62, opened 2024-05-21
What icmpv6 RA behavior to expect when same device is both ipv6 Router and Stub Router ?


alan-eero:
example: device is home gateway that allows user to enable/disable ipv6 upstream (ipv6 router), and to enable/disable device Thread border router (stub router). 

This scenario leads to the same device sending two RAs with different options content, but same link-local source address. Example:  first RA contains PIO and RDNSS. Second RA contains RIO only. 
 
This should not affect options like PIO nor RIO, because hosts should use the lifetime to deprecate the usage, but some implementations consuming RDNSS will find the second RA as an statement that recursive DNS server is no longer available so hosts using that implementation will stop using it. I believe this is a gap in that RDNSS implementation driven by the ambiguity of   https://www.rfc-editor.org/rfc/rfc6106#section-5.3.1 - however, what will be the action or expectation from SNAC to workaround / prevent these misinterpretations.

alan-eero:
FYI: @EskoDijk , @Abhayakara

Abhayakara:
This is not really a SNAC use case: the device you've described is a CE router with three interfaces, not a stub router. The three interfaces are:

- WAN
- LAN
- "Stub" network

The CE router is expected to get its internet connectivity through the WAN interface. The LAN interface and the "stub network" interface are just two different southbound interfaces, which can be configured by the router. There's no reason for the router to do any negotiation—it's the home network router. It should just publish a prefix on each link, advertise RAs on each link (assuming RAs are supported on the stub link; if not, e.g. on Thread, do what that link requires that's equivalent). 

The reason we do things so dynamically with stub networks is that there is no source of authority, but that's not really the case for a CE router. I guess if you have two CE routers both of which have a connection to infrastructure and to the stub network, you might want to find a way for them to co-operate, particularly on a constrained network like Thread.

In that case, it might make sense to implement the stub network behavior described here. The behavior on the LAN would be dictated by RFC7084, though, not by any SNAC document.

Sending two RAs seems like it would just be wrong behavior, and certainly we shouldn't explain how to do this or encourage it. If it isn't out of scope for this document, I think we should explicitly forbid it.

EskoDijk:
The current SNAC-simple draft includes in its scope a Home Gateway that includes a stub-router function. (New name for this function is SNAC router per #57). In fact all of the interoperability + usability goals still hold for this configuration. 

[...]

Issue #42, opened 2024-02-20
Add a state machine for stub network router advertisements (was: section 5.1.2.5. Stub router deprecating its on-link prefix (STATE-DEPRECATING) seems underspecified)


kiranmak:
the text in this section can be confusing when dealing with 2 usable prefixes (see section 5.2.1.4 - ULA/non-ULA).  In MCR's words: if there are two suitable prefixes, one of them will still need to be deprecated. According to Ted, this issue is orthogonal to reachability. Perhaps, we should explain that this state-machine does not talk about reachability for both the suitable prefixes. At least give forward reference to section 5.4 for complete behavior.

This will explain that we can only transition to STATE-BEGIN-ADVERTIZING and in no case we transition to STATE_USABLE

Abhayakara:
I think the issue here is not about usable prefixes, but about routes.

Suppose we have a DHCPv6 prefix, and the DHCPv6 server doesn't renew it. In this case, we probably want to stop advertising it at some point and start advertising a prefix based on our site ULA. We should talk about the timing of this switch-over.

Separately, whenever we have advertised a prefix on either link, we need to continue advertising a route to that prefix until the prefix is no longer valid on the link on which we advertised it. So if we switch from DHCPv6 prefix to ULA prefix on infrastructure, we need to be advertising routes to both for as long as both are valid.

Similarly, if we advertise a prefix on the stub network and later withdraw it because a higher-priority prefix is advertised there, we need to continue advertising a route to that prefix on the infrastructure link until the valid lifetime of the prefix we advertised has expired. I think all routers that have seen the prefix need to do this.

Then there's the case where a stub router advertising a prefix reboots. We can't assume that it's written anything to stable storage other than what prefix it's allocated for the stub network and for the infrastructure network. So when the router comes up after booting, it needs to advertise routes to these prefixes on the appropriate link until the maximum possible valid lifetime for the prefix has expired.

And finally there's the DHCPv6 case, where we got a prefix to advertise on the stub network via DHCPv6. In this case, I think we need to write the DHCPv6 prefix to stable storage when we get it, and remove it from stable storage when it expires. We should limit the lifetime we advertise on the stub network to a constant value (e.g. 30 minutes), and then if we find the DHCPv6 prefix recorded in stable storage, we advertise a route to it on infrastructure for 30 minutes after reboot. After 30 minutes, we can erase it from stable storage so that this doesn't repeat.

[...]

PR #94, opened 2024-10-07
Clarify that a SNAC router can't be a CE router

Abhayakara:
Because CE routers are infrastructure


Abhayakara:
This addresses #42

EskoDijk (approved):

EskoDijk:
@Abhayakara could we change "customer edge router" to "home gateway" (per our official terminology )? 
I see now also the RFC reference in the PR is incorrect (rfc **704**).

Abhayakara:
Closes #62

PR #133, opened 2025-07-25
Fix issue #111: Add M/O flag coordination example

### LINKED ISSUE #111: Add example where rebooted SNAC router advertises different M/O values?

Section 5.1.2.3:

> After MAX_FLAGS_COPY_TIME after reboot, the SNAC router MUST go back
> to the regular behavior defined above.  This avoids a situation where
> a SNAC router that has rebooted starts to advertise different 'M' and
> 'O' flag bit values than other SNAC routers present on the same
> link.

Could you provide an example where this scenario might take place?

Note by @EskoDijk : could be in an appendix; but adding an example of what would happen when this MUST requirement is *not* present can be also confusing to readers. (Since it won't happen when the rules in the draft are applied as expected.)

darrendukes:
## Summary
This PR addresses GitHub issue #111 by adding a concrete example to illustrate the scenario described in Section 5.1.2.3 about M/O flag coordination among SNAC routers after a reboot.

## Problem
The issue requested an example to clarify why the MAX_FLAGS_COPY_TIME mechanism is necessary to avoid situations where a rebooted SNAC router advertises different M/O flag values than other SNAC routers on the same link.

### Added New Appendix Section: "SNAC Router M/O Flag Coordination Example"

The new section includes three detailed subsections:

1. **Scenario Setup**:
   - Infrastructure Router advertising M=1, O=1 (managed configuration)
   - Two SNAC routers (A and B) both copying these values consistently
   - Normal operating state with consistent configuration signals

2. **Problem Scenario Without MAX_FLAGS_COPY_TIME Rule**:
   - Step-by-step walkthrough of what happens when SNAC Router A reboots
   - Shows how inconsistent M/O flag values would be advertised
   - Demonstrates the impact on hosts receiving conflicting Router Advertisements
   - Illustrates the specific problem the rule prevents

3. **Solution With MAX_FLAGS_COPY_TIME Rule**:
   - Step-by-step walkthrough showing how the rule solves the problem
   - Shows how consistency is maintained during the reboot process
   - Explains the transition from copying other SNAC routers to copying infrastructure router

## Technical Benefits

- **Clarifies complex protocol behavior** with concrete scenarios
- **Provides implementer guidance** on why the rule is necessary
- **Shows both problem and solution** to help understand the mechanism
- **Demonstrates real-world impact** on host configuration behavior
- **Improves specification understanding** with practical examples

## Documentation Structure
The example is placed in the appendix as suggested in the issue, maintaining the logical flow of the document while providing the requested clarification.

Closes #111

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -1124,6 +1124,75 @@
 	</t>
       </section>
       </section>
+      <section>
+	<n>SNAC Router M/O Flag Coordination Example</n>
+	<t>
+	  This section provides an example that illustrates why the MAX_FLAGS_COPY_TIME mechanism described in 
+	  <xref target="state-begin-advertising"/> is necessary to maintain consistent M/O flag values across 
+	  multiple SNAC routers on the same infrastructure link.
+	</t>
+	<section>
+	  <n>Scenario Setup</n>
+	  <t>
+	    Consider an infrastructure link with the following configuration:
+	  </t>
+	  <ul>
+	    <li>Infrastructure Router (IR): Advertises RAs with M=1, O=1 (managed configuration)</li>
+	    <li>SNAC Router A: Has been operating for hours, copying M=1, O=1 from IR</li>
+	    <li>SNAC Router B: Has been operating for hours, copying M=1, O=1 from IR</li>
+	  </ul>
+	  <t>
+	    Both SNAC routers are consistently advertising M=1, O=1 in their Router Advertisements 
+	    on the infrastructure link, maintaining consistency with the infrastructure router.
+	  </t>
+	</section>
+	<section>
+	  <n>Problem Scenario Without MAX_FLAGS_COPY_TIME Rule</n>
+	  <t>
+	    Now consider what would happen if SNAC Router A reboots and the MAX_FLAGS_COPY_TIME rule 
+	    was not present:
+	  </t>
+	  <ol>
+	    <li>SNAC Router A reboots and immediately starts advertising on the infrastructure link</li>
+	    <li>SNAC Router A has not yet received a recent RA from the Infrastructure Router (IR)</li>
+	    <li>Without the rule, SNAC Router A would default to M=0, O=0</li>
+	    <li>SNAC Router B continues advertising M=1, O=1 (copied from IR)</li>
+	    <li>Infrastructure Router continues advertising M=1, O=1</li>
+	  </ol>
+	  <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -1124,6 +1124,75 @@
...
+	  <t>
+	    Now consider what would happen if SNAC Router A reboots and the MAX_FLAGS_COPY_TIME rule 
+	    was not present:
+	  </t>
+	  <ol>
+	    <li>SNAC Router A reboots and immediately starts advertising on the infrastructure link</li>
+	    <li>SNAC Router A has not yet received a recent RA from the Infrastructure Router (IR)</li>
+	    <li>Without the rule, SNAC Router A would default to M=0, O=0</li>
+	    <li>SNAC Router B continues advertising M=1, O=1 (copied from IR)</li>
+	    <li>Infrastructure Router continues advertising M=1, O=1</li>
```
EskoDijk, line 1160: The overall scenario should be updated with IR being not available for Router A, at the time that Router A reboots: that's because
* If IR would be available, SNAC Router A would do RS and then would see the RA from IR with the flags.
* Therefore, IR needs to be not available or not responding / not connected towards Router A in this scenario.

One simple scenario to accomplish IR going away, is that we have some power plugging/unplugging event in the home that causes at the same time:
* Router A to reboot (plugged into new socket?)
* IR to power down (unplugged for the rest of the day/week? Maybe broken or getting too hot? Maybe for ISP replacement?)

In such case, back to normal operation means that after some time (MAX_FLAGS_COPY_TIME after reboot) both Router A and B will move back to M=0, O=0. Which is the correct case now the IR is gone!

Abhayakara (changes requested):
We discussed this both in Madrid and in Bangkok: maintaining consistent M&O bits isn't actually practicable. David Lamparter has some experience with previous attempts to solve this problem, and the bottom line is that it's not really possible and that instead we should rely on the host stack to do something sensible when it sees inconsistent M&O bits.

So I would suggest that we not add the section as currently proposed, and that our advice should be to just always copy the most recent M&O bits that we saw in an infrastructure RA. If infrastructure devices are advertising inconsistent M&O bits, either concurrently or over time, that's a configuration error that we can't fix.

Abhayakara:
The consensus seems to be that if we remember any valid RA from an infrastructure router, we copy the bits from the most-recently-received such RA. If we don't remember any valid RA, the bits are set to zero.

PR #136, opened 2025-07-25
Replace hardcoded /64 prefix requirement with SLAAC suitability

### LINKED ISSUE #104: Replace requirement of prefix length /64 by "suitable for IPv6 SLAAC" ?
 
> * Section 5.1.1.  Suitable On-Link Prefixes
> *  Prefix Length value is 64,
> 
> *  'L' flag bit is set and
> 
> *  either the 'A' flag bit or the 'P' flag bit 
> [I-D.ietf-6man-pio-pflag] is set, and
> 
> *  Preferred Lifetime of 30 minutes or more.
> 
> A prefix is not considered a suitable on-link prefix if the 'L' flag 
> bit is not set, or if neither the 'A' flag bit nor the 'P' flag bit 
> is set.  When the 'A' flag bit is not set, this indicates that 
> individual node addresses within the prefix are being managed using 
> DHCPv6.

I'd probably remove the reference to "/64" and simply say something
along the lines of "is suitable for IPv6 SLAAC" or the like. -- at the
end of the day, what you care is that it is suitable for SLAAC, rather
than the specific "64" value.

darrendukes:
## Summary
This PR addresses issue #104 by replacing the hardcoded "/64" prefix length requirement with language that focuses on IPv6 SLAAC suitability.

## Changes Made
- **Before:** `Prefix Length value is 64,`
- **After:** `Prefix Length is suitable for IPv6 Stateless Address Autoconfiguration (SLAAC),`

## Rationale
The reviewer feedback pointed out that we should focus on the functional requirement (SLAAC suitability) rather than hardcoding a specific prefix length value. This change:

1. **Improves clarity** - Better expresses what we actually care about (SLAAC compatibility)
2. **Allows flexibility** - While /64 is the standard for SLAAC in practice, this phrasing allows for potential future variations
3. **Maintains technical accuracy** - The requirement is functionally equivalent but more precise about the actual need

## Related Issues
Closes #104

## Review Focus
- Confirm the new language accurately captures the SLAAC suitability requirement
- Verify that the change maintains the technical intent of the original requirement

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -310,7 +310,7 @@
 	    values:
 	  </t>
 	  <ul>
-	    <li>Prefix Length value is 64,</li>
+	    <li>Prefix Length is suitable for IPv6 Stateless Address Autoconfiguration (SLAAC),</li>
             <li>'L' flag bit is set and</li>
 	    <li>either the 'A' flag bit or the 'P' flag bit <xref target="I-D.ietf-6man-pio-pflag"/> is set, and</li>
 	    <li>Preferred Lifetime of 30 minutes or more.</li>


Abhayakara:
Should add "for most use cases, the prefix length will be 64."

draft-ietf-snac-simple.xml
```diff
@@ -310,7 +310,7 @@
 	    values:
 	  </t>
 	  <ul>
-	    <li>Prefix Length value is 64,</li>
+	    <li>Prefix Length is suitable for IPv6 Stateless Address Autoconfiguration (SLAAC),</li>
```
twintersunh, line 313: Prefix length of 64 is used as that is the current prefix length supported by SLAAC [RFC4862].

EskoDijk: @twintersunh To me it looks like RFC 4862 defines SLAAC for any interface ID length of `N` bits, and it says that an implementation must not assume a certain fixed value for N. Link type documents may define their particular value of N. That said, it points out that existing architecture docs (RFCs) do define a certain value for N.   RFC 4291 calls out that this is N=64.

EskoDijk:
Agree with Ted that we should point out that in most cases (or current known cases), interface ID length (N) is 64 so that prefix length is 64 for SLAAC.

Maybe a reference to RFC 4862 or 4861 could be added for more details, or another doc that is more recent - it's a bit hard navigating the jungle of documents and requirements for this.

Abhayakara (changes requested):

draft-ietf-snac-simple.xml
```diff
@@ -310,7 +310,7 @@
 	    values:
 	  </t>
 	  <ul>
-	    <li>Prefix Length value is 64,</li>
+	    <li>Prefix Length is 64, suitable for IPv6 Stateless Address Autoconfiguration (SLAAC), consistent with current implementations and RFC4291 section 2.5.1 64bit Interface Identifiers</li>
```
darrendukes, line 313:
```suggestion
	    <li>Prefix Length is 64, suitable for IPv6 Stateless Address Autoconfiguration (SLAAC), consistent with current implementations and <xref target="RFC4291" section="2.5.1"/> 64bit Interface Identifiers</li>
```

PR #138, opened 2025-07-25
Add timeout-based prefix selection criteria for DHCPv6-PD multiple prefixes

### LINKED ISSUE #130: Which prefix type to pick if multiple prefixes can be delegated to the SNAC router?
See email https://mailarchive.ietf.org/arch/msg/snac/ciUA2nJGqG0eSa5CCpB7jUHBVjk/

This was also discussed in a SNAC IETF WG meeting, to look up if any useful guidance was given there.

Original email:

One possible situation for SNAC-simple that we haven't discussed yet, is when multiple prefixes are available for DHCPv6-PD by the infrastructure network.
For example, a CE router may be delegating from the ISP's global prefix and at the same time from its own ULA prefix.
So there may be multiple IA_PREFIX options in an IA_PD response.

The draft (5.2.3) currently only considers a single delegated prefix needs to be obtained:

  ... the SNAC router MUST attempt to acquire a prefix using DHCPv6 prefix delegation

In case of multiple, what does the SNAC router do? Pick one, or two, or more?
If there's one global prefix available and one ULA prefix, which one to pick?

  *   GUAs have the advantage of allowing hosts in the stub network to have global connectivity
  *   ULAs would only allow site-local connectivity but are potentially more stable (e.g. not impacted if the ISP decides to change the prefix delegated to the customer)
  *   Picking both has the advantage that each host on the stub network can choose its type of address(es) to configure (some hosts may support only 1 address)

If the stub network technology is in the constrained network category (like 6LoWPAN, Thread mesh) then in practice there's only limited support for advertising multiple IPv6 prefixes, and for using multiple IPv6 addresses by hosts.
In Thread networks for example, advertising an additional prefix available for SLAAC takes up precious space in the network data. And Thread nodes may limit their number of configured ULA/GUA addresses to just 1 or 2.
Also, any change in such prefixes leads to additional mesh data traffic and extra maintenance traffic is generally unwanted. From that viewpoint, advertising a stable prefix or stable subset would be desirable.

Are there any guidelines we could give for the SNAC router about what prefixes to request in this case?

darrendukes:
Addresses issue #130 by adding concrete timeout-based criteria for SNAC router prefix selection when multiple prefixes are available via DHCPv6 Prefix Delegation.

## Changes Made

This PR adds specific guidance for SNAC routers to handle scenarios where multiple prefixes (both GUA and ULA) are available for delegation. The new criteria provide concrete timeout thresholds to determine prefix selection rather than relying on vague preferences.

The intent is to focus discussion on some text to make progress on this topic

Fixes #130

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -608,18 +608,8 @@
             a SNAC router MUST select prefixes based on the following criteria, evaluated in order:
           </t>
           <ul>
-            <li>Prefix lifetime and stability: A SNAC router MUST compare the valid lifetimes and preferred lifetimes of available prefixes.
-            Prefixes with longer lifetimes (valid lifetime &gt; 24 hours) SHOULD be preferred over those with shorter lifetimes to minimize
-            renumbering events and address configuration churn on the stub network.</li>
-            <li>Prefix type preference: When prefixes have comparable lifetimes (difference &lt; 6 hours), Global Unicast Address (GUA)
-            prefixes SHOULD be preferred over ULA prefixes to enable global connectivity. However, if a ULA prefix has significantly
-            longer lifetime (&gt; 6 hours longer valid lifetime), it MAY be preferred for stability reasons.</li>
-            <li>Single prefix constraint: For constrained stub networks (e.g., 6LoWPAN, Thread mesh networks) that have limited
-            support for multiple prefixes, a SNAC router MUST select only the single best prefix according to the above criteria
-            to minimize network overhead, maintenance traffic, and memory consumption.</li>
-            <li>Multiple prefix support: For stub networks that can efficiently support multiple prefixes, a SNAC router MAY
-            accept and advertise both the best GUA and best ULA prefix (if both meet minimum lifetime requirements of valid lifetime &gt; 2 hours),
-            allowing hosts to choose their preferred address type.</li>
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -603,6 +603,29 @@
...
+          <ul>
+            <li>Prefix lifetime and stability: A SNAC router MUST compare the valid lifetimes and preferred lifetimes of available prefixes.
+            Prefixes with longer lifetimes (valid lifetime &gt; 24 hours) SHOULD be preferred over those with shorter lifetimes to minimize
+            renumbering events and address configuration churn on the stub network.</li>
+            <li>Prefix type preference: When prefixes have comparable lifetimes (difference &lt; 6 hours), Global Unicast Address (GUA)
+            prefixes SHOULD be preferred over ULA prefixes to enable global connectivity. However, if a ULA prefix has significantly
+            longer lifetime (&gt; 6 hours longer valid lifetime), it MAY be preferred for stability reasons.</li>
+            <li>Single prefix constraint: For constrained stub networks (e.g., 6LoWPAN, Thread mesh networks) that have limited
+            support for multiple prefixes, a SNAC router MUST select only the single best prefix according to the above criteria
+            to minimize network overhead, maintenance traffic, and memory consumption.</li>
```
darrendukes, line 619: Are there cases where the stub network technology necessitates use of ULA over GUA? If so the text below is wrong.

```suggestion
            <li>Single prefix constraint: For constrained stub networks (e.g., 6LoWPAN, Thread mesh networks) that have limited
            support for multiple prefixes, a SNAC router MUST select only the single best prefix according to the above criteria and 
            to minimize network overhead, maintenance traffic, and memory consumption.</li>
```

EskoDijk: No case that I know of.

There can be managed deployments in which ULA is required for the stub network nodes, but this will be administratively configured and not SNAC-autoconfigured.

PR #141, opened 2025-09-09
Handle M and O bits per consensus in stub network RAs - Fix #111

### LINKED ISSUE #111: Add example where rebooted SNAC router advertises different M/O values?

darrendukes:
Implements working group consensus for handling M and O bits in stub network Router Advertisements:

- If the SNAC router has received and remembers any valid (not expired) Router Advertisement (RA) from an infrastructure router, it MUST copy the M and O bits from the most-recently-received such RA into advertisements on the stub network
- If no valid RA is available (none received or all expired), the SNAC router MUST set the M and O bits to zero in advertisements on the stub network

This change replaces the previous behavior where M and O bits were always set to zero.

Fixes #111

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -1211,9 +1211,9 @@
         given in Section 4.2 of <xref target="RFC4861"/>:
       </t>
 	<ul>
-          <li> Router Lifetime: The SNAC router can be a default router on the stub network (see xref target="snac-reachability" TBD fixref).</li>
-          <li> SNAC routers do not provide DHCP service on the stub network. Therefore, the 'M' and 'O' flag bits MUST be zero.</li>
-	  <li> The 'SNAC router' flag bit xref target="I-D.ietf-6man-snac-router-bit" TBD waitRef MUST be 0.</li>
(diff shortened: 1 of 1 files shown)


EskoDijk:
@darrendukes I think the PR here changes the wrong part - see original issue. It was about M/O flags advertised on the AIL, not the M/O flags for the stub network which are always '0' - since the SNAC router doesn't provide DHCP on the stub network, but rather it relies on SLAAC.

EskoDijk:
Proposing to close/reject this PR. There's no specified requirement to copy M/O values from AIL to stub; rather the original text remains correct: on the stub, M=0 and O=0 given that the SNAC router provides no DHCPv6 (server/relay) functions.

PR #143, opened 2025-09-10
[editorial] apply consistent caps in terms; and minor editorial fixes

### LINKED ISSUE #49: Use consistent caps for terms (editorial)
Just a reminder for future editing: it would be nice if defined terms (taken from glossary) use caps consistently.
E.g. we may have 
* "adjacent infrastructure link" vs "Adjacent Infrastructure Link"
* "ULA link prefix" vs "ULA Link Prefix"

The request here is that if we use lowercase for one term, preferably also use lowercase for other terms - have a default policy on a preferred form, if there's no other reason to pick left or right.

EskoDijk:
This sets the caps usage to mostly lowercase as agreed in issue #49. This also follows the most common caps usage for terms that are imported from other RFCs. Caps use in section titles is kept (since using Caps in section titles is typical.)  Close #49


Abhayakara (approved):
This looks fine.

jwhui (approved):

PR #145, opened 2025-09-11
[editorial] avoid use of 'we' and 'us'

### LINKED ISSUE #128: Avoid the use of "we"
Avoid the use of "we" in the document as it is unclear whether the "we" term is for the authors, the SNAC WG, the IETF community (after the IETF last call).

EskoDijk:
This replaces the 'we' and 'us' wording, which wasn't so clear who it referred to, with other phrasing e.g. 'SNAC router' or unspecified form 'one assumes that ...' or passive form.
Closes #128

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -801,8 +801,8 @@
 	<section>
 	  <name>Providing discoverability of adjacent infrastructure hosts on the stub network</name>
 	  <t>
-	    Hosts on the stub network may need to discover hosts on the AIL, or on the stub network.  In
-	    an IoT use case for example, there might be a light switch on the stub network which needs to be able to
+	    Hosts on the stub network may need to discover hosts on the AIL, or on the stub network. In
+	    an IoT use case, for example, there might be a light switch on the stub network which needs to be able to
 	    actuate a light bulb connected to the AIL. In order to know where to send the actuation
 	    messages, the light switch will need to be able to discover the light bulb's address somehow.</t>
 	  <t>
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -971,8 +971,8 @@
     <section>
       <name>Services Provided by SNAC routers</name>
       <t>
-	In order to provide network access, SNAC routers must provide some network services to the stub network. We have previously
-	discussed the following services:</t>
+	In order to provide network access, SNAC routers must provide some network services to the stub network. In this document
+	the following services were discussed:</t>
```
Abhayakara, line 975: "have been discussed?

draft-ietf-snac-simple.xml
```diff
@@ -1117,18 +1117,18 @@
...
 	  <t>
-	    In such scenarios we have no way to actually know based on the network configuration what the operator's intention was. An operator
+	    In such scenarios there is no way to actually know based on the network configuration what the operator's intention was. An operator
 	    that sees a problem with this can react by implementing RA guard or by blocking unknown source addresses at the router, and in so
 	    doing they would be expressing their intention. This configuration would not cause any new problem: a host that could communicate would
 	    still be able to communicate, and a host that could not communicate would not become able to communicate.
 	  </t>
 	  <t>
-	    The one scenario where we could actually see a communication problem is when there is a GUA prefix advertised by
+	    The one scenario where actually a communication problem can be expected is when there is a GUA prefix advertised by
```
Abhayakara, line 1131: "The one scenario where a communication problem can actually be expected..."

draft-ietf-snac-simple.xml
```diff
@@ -1117,18 +1117,18 @@
 	  </t>
 	  <t>
 	    However, this can only happen in practice if the host did not receive an address from DHCPv6.  In this case, the host
-	    would not be able to communicate anyway. So the behavior we might see here is that a series of IP packets with an
+	    would not be able to communicate anyway. So the potential behavior here is that a series of IP packets with an
```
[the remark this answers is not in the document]
EskoDijk: Changing to:  The problem that might occur here is that a series of IP packets with an unexpected source address are sent to a device on another link, and the device would be unable to send a response.

PR #150, opened 2025-09-30
Sync message and constant names with RFC4861

### LINKED ISSUE #98: Sync constant names with RFC4861, reword text as appropriate
Currently e.g. RA_BEACON_INTERVAL is the same as MinRtrAdvInterval and MaxRtrAdvInterval. This is true of other constants as well. We should use the same constant names when we can, and just override the defaults as appropriate. Also don't use the term "Beacon." An RA beacon is a multicast RA. Update the text to clarify that.

EskoDijk:
Close #98


EskoDijk:
(Interim)
Ted: section line 611-615 seems not really needed (anymore). For Thread, remembering prefix isn't needed (using fixed PIO prefix).

EskoDijk:
PR now updated based on SNAC-interim discussion: use values MinRtrAdvInterval/MaxRtrAdvInterval directly as the constants/parameters.  Also as a side-effect this fixes the protocol for the case that a SNAC router is configured with *other* values than the defaults, for these.

PR #154, opened 2025-09-30
updated MO advertisement, removed timer

darrendukes:
Resolve Issue #111 
Updated MO advertisement as per last consensus. Removed MAX_FLAGS_COPY_TIME


darrendukes:
reviewed in Interim and text was correct as per Ted and Esko

PR #172, opened 2026-01-02
Clarify DNS terms incl 'stub resolver' and make reqs normative

### LINKED ISSUE #142: Define / clarify "stub resolver" and avoid requirements on entity "network"
Following texts mention "stub resolver" which is not defined in the glossary.

>  	  A stub network that services a Wi-Fi stub network SHOULD provide DNS64 translation: hosts on the stub network cannot be
> 	  assumed to be able to do DNS64 synthesis in the stub resolver. In this case the DNS resolver on the SNAC router MUST honor
> 	  the 'CD' and 'DO' flag bits if received in a request, since this indicates that the stub resolver on the requestor intends to do
> 	  DNSSEC validation. In this case, the resolver on the SNAC router MUST NOT perform DNS64 synthesis.
> 

> 	<dt>Legacy Browsing Domains:</dt><dd>The stub resolver must advertise legacy browsing domains for each AIL,
> 	  for the zone that is maintained by its SRP service, and in addition must list the legacy browsing domains provided
> 	  by the infrastructure network, if any.</dd>
> 

* We should define the term, or use a more clear like "DNS resolver" or "the DNS resolver in the SNAC router" if we mean that.
* Also, there's a requirement on the entity "stub network" in the top text. Shouldn't that be on the SNAC router instead?
* Also the bottom text has a "must" (no caps) requirement, can this be MUST? Can this requirement be also put on the SNAC router instead? (i.e. the DNS resolver in the SNAC router that's mentioned in Section 8 higher up.)

EskoDijk:
This clarifies use of some DNS terms, by pointing to RFC 9499; also it makes some requirements normative as discussed in an interim call.

Closes #142

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -248,7 +248,7 @@
 	<dt>Stub Router:</dt>
 	<dd>A router that provides connectivity between a stub network and an infrastructure network. A stub router may also provide
 	  connectivity between other networks: the term "stub router" refers specifically to its role in providing connectivity to
-	  a stub network. For example, a CE Router may provide connectivity between a provider network (WAN) and a  network
+	  a stub network. For example, a CE Router may provide connectivity between a provider network (WAN) and a network
 	  (LAN), while at the same time providing connectivity between the LAN and a stub network. What distinguishes the LAN from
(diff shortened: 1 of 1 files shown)


EskoDijk:
@jwhui  Did you want to review or approve this item? I wasn't sure!

jwhui (approved):

PR #177, opened 2026-01-23
Fix ULA site/link prefix renumbering requirements and describe exceptions

### LINKED ISSUE #158: Unclear requirement on regenerating ULA site prefix
There's this text in the ULA Site Prefix section:

>  Any ULA link prefixes allocated by a SNAC router SHOULD be maintained across reboots, and SHOULD remain stable over time. (TBD:
> mention the SHOULD exception cases) However, for privacy reasons, a SNAC router that roams from network to network
> SHOULD allocate a different ULA link prefix each time it connects to a different infrastructure network, unless configured to
> behave otherwise.

If only the ULA link prefix changes while the ULA site prefix remains the same, there's no privacy.
Do we mean that the ULA site prefix is also regenerated on a different infra network?
Or did we mean that only the  16-bit "Subnet ID" changes?

Also the "TBD exception case" is Thread for example: single random Global ID + Subnet ID based on random Extended PAN ID of the stub network. This won't change when roaming.

EskoDijk:
Close #158

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -624,16 +624,15 @@
 	    (<xref target="RFC4193"/>) are randomly allocated prefixes. A SNAC router MUST allocate a single ULA site prefix for use in
 	    providing on-link prefixes to the stub network and the adjacent infrastructure link as described in <xref target="state-begin-advertising"/>.
 	  </t><t>
-	    Any ULA link prefixes allocated by a SNAC router SHOULD be maintained across reboots, and SHOULD remain stable over time.
-        An exception is the following: for privacy reasons, a SNAC router that roams from one network to
-        another network and can detect that its AIL changed, SHOULD allocate a different ULA site prefix each time
-        it connects to a different AIL.
-        On the latter recommendation there are two possible exceptions:
+	    Any ULA link prefixes allocated by a SNAC router SHOULD persist across reboots, and SHOULD remain stable over time.
+        An exception to both these requirements is the following: for privacy reasons, a SNAC router that can detect
+        that its AIL changed, SHOULD allocate a different ULA site prefix each time it connects to a different AIL.
+        On the latter requirement there are two possible exceptions:
 	  </t>
         <ol>
-            <li>The SNAC router remains connected to the same stub network and the ULA site prefix value is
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -624,11 +624,17 @@
...
+	    Any ULA link prefixes allocated by a SNAC router SHOULD be maintained across reboots, and SHOULD remain stable over time.
+        An exception is the following: for privacy reasons, a SNAC router that roams from one network to
+        another network and can detect that its AIL changed, SHOULD allocate a different ULA site prefix each time
+        it connects to a different AIL.
+        On the latter recommendation there are two possible exceptions:
 	  </t>
+        <ol>
+            <li>The SNAC router remains connected to the same stub network and the ULA site prefix value is
+                generated based on properties of the stub network (such as configuration data).</li>
+           <li>The SNAC router is configured to behave otherwise.</li>
```
Abhayakara: configured by the user or operator

draft-ietf-snac-simple.xml
```diff
@@ -624,11 +624,17 @@
 	    (<xref target="RFC4193"/>) are randomly allocated prefixes. A SNAC router MUST allocate a single ULA site prefix for use in
 	    providing on-link prefixes to the stub network and the adjacent infrastructure link as described in <xref target="state-begin-advertising"/>.
 	  </t><t>
-	    Any ULA link prefixes allocated by a SNAC router SHOULD be maintained across reboots, and SHOULD remain stable over time. (TBD:
-            mention the SHOULD exception cases) However, for privacy reasons, a SNAC router that roams from network to network
-            SHOULD allocate a different ULA link prefix each time it connects to a different infrastructure network, unless configured to
-            behave otherwise.
+	    Any ULA link prefixes allocated by a SNAC router SHOULD be maintained across reboots, and SHOULD remain stable over time.
```
EskoDijk: Discussion: also exception cases for the SHOULDs here.

draft-ietf-snac-simple.xml
EskoDijk: done in new commit

draft-ietf-snac-simple.xml
EskoDijk: checked text, all exceptions were there -- now simplified wording a bit. (Upcoming commit)

PR #179, opened 2026-04-13
Add normative NAT64 and DNS64 references.

EskoDijk:
Adding references based on Med's review comment: https://mailarchive.ietf.org/arch/msg/snac/tF9jVVVF9pZPmBHZYguPkXaS29A/

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -260,7 +260,7 @@
 	  ULA link prefixes to the stub network and the adjacent infrastructure link.</dd>
 	<dt>ULA Link Prefix:</dt>
 	<dd>A Unique Local Address /64 prefix allocated from the ULA site prefix. SNAC routers can use ULA Link prefixes to provide
-	addressability on the stub network and/or adjacent infrastructure link as needed. If a SNAC router is doing NAT64, the NAT64
+	addressability on the stub network and/or adjacent infrastructure link as needed. If a SNAC router is doing NAT64 <xref target="RFC6146"/>, the NAT64
 	prefix is also a ULA link prefix. A total of 65,536 ULA link prefixes can be allocated from the ULA site prefix.</dd>
       </dl>
     </section>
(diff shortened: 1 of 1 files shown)


boucadair (approved):
Thanks @EskoDijk

EskoDijk:
Discussion followed on the mailing list thread: it was (I think) proposed to remove the SNAC router doing DNS64, rather, let the DNS client (stub network host) do any DNS IPv4/IPv6 translation itself.
Currently it's listed for the Wi-Fi case that the SNAC router has to do DNS64.

jwhui (approved):

draft-ietf-snac-simple.xml
```diff
@@ -980,7 +980,7 @@
 	  uses its own NAT64 prefix—there is no common NAT64 prefix. The reason for this is that NAT64 translation is not stateless,
 	  and is tied to the SNAC router&apos;s IPv4 address. Therefore, each NAT64 egress is not equivalent.
 	</t><t>
-	  A SNAC router that services a Wi-Fi stub network SHOULD provide DNS64 translation: IPv6 hosts on the stub network cannot be
+	  A SNAC router that services a Wi-Fi stub network SHOULD provide DNS64 <xref target="RFC6147"/> translation: IPv6 hosts on the stub network cannot be
```
Abhayakara: Yeah, I think we need to decide what to do here. There's a whole discussion going on on DNSOP right now that is sort of "DNS64 considered harmful" versus "best is the enemy of good enough." I think most devices we care about know how to do local synthesis at this point, so personally I would rather require PREF64 where applicable, and "the equivalent" where not, and let the host sort it out.

EskoDijk:
@Abhayakara  Now rephrased to remove the DNS64 requirement. As such, DNS64 is now informative. Per discussion it's expected that the stub network host performs the synthesis if it needs to.

I'll merge this by tomorrow in case no further comments received!

draft-ietf-snac-simple.xml
EskoDijk: see futher below - now rephrased.

Issue #180, opened 2026-04-13
Handling multiple AIL-interfaces in a single SNAC router


EskoDijk:
So far, the text was always written (at multiple places) for  a SNAC router that could contain multiple AIL interfaces. For example, both Ethernet and Wi-Fi.

Although this is a multi-AIL situation, it differs from the one discussed in #139 because in that case, different SNAC routers connect to different sets of AILs, leading to potential routing and discovery problems.

For this case of "multiple AIL interfaces on 1 SNAC router" we need to decide whether to keep that in scope or not. 

What are benefits of limiting ourselves to a single active AIL interface?  I.e. interface being active means active for SNAC reachability and discoverability.
* spec becomes simpler
* we can avoid a solution for how the 2 or more mDNS ".local" domains can be merged into a single DNS domain for stub network clients to use
* we can avoid potential problems when the 2 distinct AILs are (without the SNAC router knowing this....) mDNS-proxied at another device like a CE router.  CE routers seem to do this sometimes to proxy mDNS between Wi-Fi and Ethernet links (?)

Disadvantages of limiting to a single SNAC-active AIL interface:
* in case of 2 AIL interfaces on the SNAC router, one class of device is not being discovered and can't be reached (e.g. we miss the Wi-Fi light bulb if Ethernet is chosen as the AIL).
* it may be unclear to users which AIL gets selected/chosen or how to influence that.
* possibly unpredictable or counter-intuitive network behavior for users.

Abhayakara:
One thing to bear in mind is that if we don't solve the multi-AIL problem in this document that doesn't mean that vendors can't correctly solve it—it just means that in doing so they are going beyond what is specified here. IOW the disadvantages you list assume that the vendor provides multiple AILs that are separate multicast domains (unusual) _and_ that they don't do anything to make it work correctly.

EskoDijk:
PR merged, closing issue.

PR #181, opened 2026-04-14
Change multiple AIL interface support to single AIL only, consistently.

EskoDijk:
Related to issue #180.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -826,10 +826,8 @@
 	    By implication, this means that a SNAC router MUST provide a DNS resolver. In addition, a SNAC router MUST provide
 	    a DNS zone for the AIL it is attached to, and MUST list this zone in the list of default browsing domains
 	    as defined in <xref target="RFC6763" section="11" sectionFormat="of"/>.
-		A SNAC router MUST enable a Discovery Proxy on its AIL that translates between the name of above DNS zone and
-		the local domain on the AIL.
-	    [[WG TBD: do we need to say how the DNS zone is named?
-		  Or refer to the Advertising Proxy doc and have that doc say how it is named? Why would the name matter?]]</t>
+		A SNAC router MUST provide a Discovery Proxy that operates with this DNS zone.
+	    [[WG TBD: issue #183 is to define name or name requirements.]]</t>
 	  <t>
 	    The SNAC router MUST also maintain an SRP registrar and use registrations made through that SRP registrar to populate
 	    a DNS zone which is advertised as a default browsing domain, as defined above.
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -753,9 +753,9 @@
 	  case, if the infrastructure is not capable of routing between these two AILs, a packet which could have been delivered by
 	  another SNAC router will be lost by the SNAC router that received it.</t>
 	<t>
-	  Consequently, SNAC routers SHOULD be configurable to not advertise themselves as default routers on the stub network. SNAC
-	  routers SHOULD be configurable to explicitly advertise its AIL prefixes on the stub network even if they are advertising as a
-	  default router. The mechanisms by which such configuration can be accomplished are out of scope for this document.
+	  Consequently, a SNAC router SHOULD be configurable to not advertise itself as a default router on the stub network.
+	  A SNAC router SHOULD be configurable to explicitly advertise AIL prefixes on the stub network even if it is advertising as a
+	  default router on the stub network. The mechanisms by which such configuration can be accomplished are out of scope for this document.
 	</t>
 	<t>
 	  It is also possible that SNAC routers for more than one stub network may be connected to the same AIL.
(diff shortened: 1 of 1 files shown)


jwhui (approved):

draft-ietf-snac-simple.xml
```diff
@@ -754,14 +754,14 @@
 	  another SNAC router will be lost by the SNAC router that received it.</t>
 	<t>
 	  Consequently, SNAC routers SHOULD be configurable to not advertise themselves as default routers on the stub network. SNAC
-	  routers SHOULD be configurable to explicitly advertise AIL prefixes on the stub network even if they are advertising as a
+	  routers SHOULD be configurable to explicitly advertise its AIL prefixes on the stub network even if they are advertising as a
```
Abhayakara: This could be read as "the prefixes it advertises as on-link on the AIL." I think it would be clearer to say "prefixes that are being advertised as on-link on its AIL" as previously, even though that's longer. Or just don't add its—I don't think it's necessary.

draft-ietf-snac-simple.xml
```diff
@@ -824,11 +824,12 @@
...
+	    a DNS zone for the AIL it is attached to, and MUST list this zone in the list of default browsing domains
 	    as defined in <xref target="RFC6763" section="11" sectionFormat="of"/>.
-		For each of these zones, a SNAC router MUST enable a Discovery Proxy on the corresponding AIL.
-	    [[WG TBD: do we need to say how these zones are named? Or refer to the Advertising Proxy doc and have that doc say how they are named? Why would the names matter?
-		   Do we also need to say a client can *query* on the domain "default.service.arpa" and get aggregated results of all AILs back?]]</t>
+		A SNAC router MUST enable a Discovery Proxy on its AIL that translates between the name of above DNS zone and the "local." domain on the AIL.
+	    [[WG TBD: do we need to say how the DNS zone is named? Or refer to the Advertising Proxy doc and have that doc say how it is named? Why would the name matter?
+		   Do we also need to say a client can *query* on the domain "default.service.arpa" and get aggregated results of all AILs back?
+		   This is very useful for constrained DNS clients in the stub network.]]</t>
```
Abhayakara: The text as written should work for the general case. RFC6763 already says how to do this, and if a SNAC router behaves as specified here and in RFC6763, conforming DNSSD queriers will be able to do service discovery. The benefit of allowing queries in default.service.arpa, as we do for Thread, is that the client doesn't have to enumerate the default browsing domains, but this is a constrained-node-specific problem that we don't actually _have_ to solve here. Given that the DNSSD working group is hopefully going to be working on a document that describes this more generally, and given that we have a future snac-complicated document, I think it's okay to just not specify anything new here. In particular, I don't think we should require the use of .local here. Let's leave that for the DNSSD service document.

Abhayakara (changes requested):
I've made a couple of comments here that I think need to be addressed, but this is mostly ready.

draft-ietf-snac-simple.xml
EskoDijk: Yes, agree with your point -- a future document could define the "shortcut" that we use in Thread. Classic RFC 6763 querying is available in any case. So I could update or remove the TBD here.

There's still the "zone name" question in here that I think would need some discussion.

draft-ietf-snac-simple.xml
EskoDijk: Oops, not sure how that word got in. I'll update. One thing I'll do is change the normative requirement on the SNAC router to singular, because the plural requirements (i.e. on a group of SNAC routers) are more difficult to parse , and may be wrongly interpreted as a true requirement on the group, while it's a requirement on a single implementation in reality.

draft-ietf-snac-simple.xml
Abhayakara: "home.arpa"? :)

But you raise a good point: we really don't want them to just pick some random domain. I think explaining how default.service.arpa is used on Thread is out of scope, but we could chose a zone under service.arpa (snac.service.arpa) and say that subdomains of that zone MUST be used if no delegation from the root exists.

EskoDijk: One thing I don't understand: 

> I don't think we should require the use of .local here.

It's not a new requirement I think; RFC 8766 is fully written towards proxying between .local and another domain. It would be strange not to call the first domain ".local" ?
Saying "the local domain of the AIL" is an alternative but sounds equivalent.

EskoDijk:
> I've made a couple of comments here that I think need to be addressed, but this is mostly ready.

Now updated based on your comments!  As a second commit to be able to compare.

draft-ietf-snac-simple.xml
Abhayakara: OK, I see the disconnect. We don't need to mention ".local" at all because that's covered in RFC8766. I thought you meant doing discovery proxy queries with .local as the domain in which the query is done.

There are actually two things here:

1. Discovery Proxy is going to wind up discovering what the advertising proxy is advertising, because that's going to show up in .local
2. We do actually need to specify a domain to use. Could be default.service.arpa. Given that in the simplest case all we are going to do is require discovery proxy, what we actually need to make sure of is that we don't accidentally censor the infrastructure, which may also be publishing default browsing domains. So we need to specify a new domain to add to the default browsing domain list, and additionally require the snac router to (by default) also include the default browsing domains discovered on the AIL.

I don't think we want to do more than that, e.g. we do not want to specify a single domain that merges answers from all these other domains. That's properly work for the DNSSD server work in the DNSSD working group.

evyncke: Indeed rfc6762 section 3 clearly forbids to use .local :(

So much I am not a fan of home.arpa https://datatracker.ietf.org/doc/html/rfc8375#section-8 would allow it

EskoDijk: Discussion: "delegation" of name from ISP -> not in our scope now.  
* "default.service.arpa" -> good option? 
   * update the IANA registration for this name -> also for queries.
   * simplest solution.
* Or it could be a subdomain like "<myname>.snac.service.arpa" ? 
* Or subdomains of default.service.arpa 

For further naming discussion see #183

draft-ietf-snac-simple.xml
```diff
@@ -824,11 +824,12 @@
 	    provided on the stub network, as described in <xref target="RFC6763" section="11"/>.</t>
 	  <t>
 	    By implication, this means that a SNAC router MUST provide a DNS resolver. In addition, a SNAC router MUST provide
-	    a DNS zone for each AIL it is attached to, and MUST list all these zones in the list of default browsing domains
+	    a DNS zone for the AIL it is attached to, and MUST list this zone in the list of default browsing domains
 	    as defined in <xref target="RFC6763" section="11" sectionFormat="of"/>.
-		For each of these zones, a SNAC router MUST enable a Discovery Proxy on the corresponding AIL.
-	    [[WG TBD: do we need to say how these zones are named? Or refer to the Advertising Proxy doc and have that doc say how they are named? Why would the names matter?
-		   Do we also need to say a client can *query* on the domain "default.service.arpa" and get aggregated results of all AILs back?]]</t>
+		A SNAC router MUST enable a Discovery Proxy on its AIL that translates between the name of above DNS zone and
```
EskoDijk:
```suggestion
		A SNAC router MUST provide a Discovery Proxy to allow ... on its AIL.
```

Don't mention what DiscProxy RFC already says.

EskoDijk: New text will also mention that the Discovery Proxy uses the aforementioned DNS zone; as the original text did. Exact wording of that can be changed post-WGLC.

PR #184, opened 2026-04-16
Generalize text around DNSSEC; don't make normative requirements on stub network host for DNSSEC

### LINKED ISSUE #176: Security considerations: require a SNAC router to perform DNSSEC validation?
We discussed the possibility that secure DNS (over TLS) is used to the SNAC router, which then "downgrades" to unsecure DNS to reach the infrastructure DNS resolver. See PR #175 and issue #165 .

Would it help to recommend or require the SNAC router to perform DNSSEC validation, in case unsecure DNS was used to the infrastructure resolver?

Pros:
* some tamper/spoofing security - even when using unsecure DNS
* validation is done even when constrained stub host doesn't implement it

Cons:
* latency
* code size
* stub network clients may come to "rely" on this DNSSEC validation and be made less secure themselves

EskoDijk:
Also introduce 'upstream' resolver to cover all of internet, ISP and home/AIL based DNS resolvers.

Per discussion in issue #176 ; close #176

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -1119,27 +1119,32 @@
         A SNAC router MUST support DNS-over-TLS as specified in <xref target="RFC9665" section="7" sectionFormat="of"/>
         for any DNS unicast communication with hosts (acting as DNS clients) in the stub network.
         This provides opportunistic privacy for SRP Updates as well as for DNS queries.
+		See <xref target="RFC7435"/> for the "opportunistic security" concept and
+	    <xref target="RFC7858" section="3.1" sectionFormat="of"/> for the opportunistic privacy profile as defined for DNS-over-TLS.
+	    This prevents eavesdroppers on the stub network that are not able to intercept the TLS connection from determining
+	    what queries are being made. However, there is no protection against interception on the stub network other than
+	    that it may be difficult to accomplish.
+        Support for non-opportunistic DNS-over-TLS is out of scope for this document.
+      </t><t>
+        There is also no guarantee that privacy-preserving DNS service will be available on the AIL.
+		When such DNS service is present, the SNAC router SHOULD use it, if it is technically capable of doing so.
+		However, there is no way to signal on the stub network that this is being done.
+		Given that the privacy profile supported here is opportunistic, the DNS client has no guarantee of privacy.
+		Aggregation of queries originating from the stub network by the SNAC router provides some degree of anonymization,
+		but in any case because the TLS server's certificate is not being validated by the client, the client should
+		not assume it is getting more privacy than an unauthenticated TLS connection can deliver.
       </t>
       <t>
         If a SNAC router receives a DNS query protected by TLS from a stub network host, it may happen that the query
-        is forwarded without security to the DNS resolver configured in the infrastructure network.
-        For example, if the upstream DNS resolver only supports unsecured DNS. In such cases, the SNAC router falls
-        back to sending unsecured DNS queries, in order to meet the SNAC interoperability and usability goals.
-      </t>
-      <t>
-        This fallback from DNS-over-TLS to unsecured DNS removes the confidentiality and integrity protections provided
-        by TLS and exposes DNS queries and responses to on-path observers. An attacker on the path between the SNAC router
-        and the infrastructure resolver can then eavesdrop on DNS traffic, perform traffic analysis, and inject or modify DNS
-        responses (for example, to redirect stub hosts to attacker-controlled services). To mitigate these risks,
-        SNAC routers MUST prefer upstream DNS resolvers that support DNS-over-TLS or another form of secured DNS
-        when such resolvers are available, and SHOULD log or otherwise surface to device/network administrators when
-        encrypted queries from stub hosts are downgraded to unsecured DNS.
+        is forwarded without TLS security to the upstream DNS resolver as configured by the infrastructure network.
+	    This enables an on-path eavesdropper in the infrastructure network to observe any DNS queries that the SNAC
+	    router forwards to the upstream DNS resolver, or an on-path attacker to spoof DNS responses to such queries.
       </t>
       <t>
         Stub network hosts that implement DNSSEC <xref target="RFC9364"/> can perform DNSSEC validation of DNS responses
-        for non-local domain queries.
-        Note that DNS records for infrastructure-network-local hosts or services (e.g. records registered via SRP
-        or discovered by the SNAC router using mDNS) do not support DNSSEC protection.
+        received from an upstream DNS resolver.
+	    Note that for locally-served zones {{RFC6303}} DNSSEC validation cannot work and is expected
+	    to be skipped by the SNAC router.
       </t>
     </section>
   </middle>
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -1055,22 +1055,21 @@
       <t>
         If a SNAC router receives a DNS query protected by TLS from a stub network host, it may happen that the query
         is forwarded without security to the DNS resolver configured in the infrastructure network.
-        One specific example when this happens is when the configured DNS resolver is inside a CE router and it only
-        supports unsecured DNS. In such cases, the SNAC router is expected to fall back to sending unsecured DNS queries
-        in order to meet the SNAC interoperability and usability goals.
+        For example, if the upstream DNS resolver only supports unsecured DNS. In such cases, the SNAC router falls
+        back to sending unsecured DNS queries, in order to meet the SNAC interoperability and usability goals.
```
Abhayakara: I think we should not say "unsecured DNS" because that implies that the DNS-over-TLS that's provided by the SNAC router is secure, and it's not.

What I would suggest for this paragraph (the part you've replaced) is:

Support for non-opportunistic DNS-over-TLS is out of scope for this document. A SNAC router as described in this specification that receives a DNS query using DNS-over-TLS is providing opportunistic security, as described in {{RFC7435}} and in Section 3.1 of {{RFC7858}}. This prevents eavesdroppers that are not able to intercept the TLS connection from determining what queries are being made. However, there is no protection against interception other than that it may be difficult to accomplish.

There is also no guarantee that privacy-preserving DNS service will be available on the infrastructure link. When such service is present, the SNAC router SHOULD use it, if it is technically capable of doing so. However, there is no way to signal on the stub network that this is being done. Given that the privacy profile supported here is opportunistic, the client has no guarantee of privacy. Aggregation of queries from the stub link provides some degree of anonymization, but in any case because the TLS certificate is not being validated, the client could not be assuming that it's getting more privacy than this approach actually delivers.

draft-ietf-snac-simple.xml
```diff
@@ -1055,22 +1055,21 @@
...
       </t>
       <t>
         This fallback from DNS-over-TLS to unsecured DNS removes the confidentiality and integrity protections provided
         by TLS and exposes DNS queries and responses to on-path observers. An attacker on the path between the SNAC router
         and the infrastructure resolver can then eavesdrop on DNS traffic, perform traffic analysis, and inject or modify DNS
         responses (for example, to redirect stub hosts to attacker-controlled services). To mitigate these risks,
-        SNAC routers MUST prefer infrastructure network resolvers that support DNS-over-TLS or another form of secured DNS
+        SNAC routers MUST prefer upstream DNS resolvers that support DNS-over-TLS or another form of secured DNS
         when such resolvers are available, and SHOULD log or otherwise surface to device/network administrators when
```
Abhayakara: I would delete this entire paragraph and consider the second paragraph in my previous comment as its replacement.

draft-ietf-snac-simple.xml
```diff
@@ -1055,22 +1055,21 @@
...
         encrypted queries from stub hosts are downgraded to unsecured DNS.
       </t>
       <t>
-        Stub network hosts are also RECOMMENDED to perform DNSSEC validation where feasible, in order to detect spoofed DNS responses
-        which may result from DNS queries and responses being forwarded in unsecured DNS messages.
+        Stub network hosts that implement DNSSEC <xref target="RFC9364"/> can perform DNSSEC validation of DNS responses
+        for non-local domain queries.
```
Abhayakara: the term "non-local" here doesn't mean anything specific, and so this clause should be deleted. That is, just delete "for non-local domain queries."

Note that for locally-served domains, it's assumed that DNSSEC validation would find the domains to be nonexistent. If this is what's meant to be excluded by "non-local", you could add a sentence: "Note that for locally-served zones, as described in {{RFC6303}}, DNSSEC validation can't work and is expected to be skipped."

Abhayakara (changes requested):

draft-ietf-snac-simple.xml
EskoDijk: In the upcoming commit, I've integrated this text - but modified and placed higher up (where we introduce 'opportunistic privacy' already).

draft-ietf-snac-simple.xml
EskoDijk: Paragraph is deleted in upcoming commit; with some minor items moved to other paragraphs.  (Things I feel are important the WG must review before we decide to delete it.)

draft-ietf-snac-simple.xml
EskoDijk: Good suggestion, I've added the proposed line in the new commit and removed "non-local".

EskoDijk:
@Abhayakara Now addressed the comments, and my own concerns, with the PR rebased.

draft-ietf-snac-simple.xml
```diff
@@ -1081,34 +1080,38 @@
...
+	    that it may be difficult to accomplish.
+        Support for non-opportunistic DNS-over-TLS is out of scope for this document.
+      </t><t>
+        There is also no guarantee that privacy-preserving DNS service will be available on the AIL.
+		When such DNS service is present, the SNAC router SHOULD use it, if it is technically capable of doing so.
+		However, there is no way to signal on the stub network that this is being done.
+		Given that the privacy profile supported here is opportunistic, the DNS client has no guarantee of privacy.
+		Aggregation of queries originating from the stub network by the SNAC router provides some degree of anonymization,
+		but in any case because the TLS server's certificate is not being validated by the client, the client should
+		not assume it is getting more privacy than an unauthenticated TLS connection can deliver.
```
Abhayakara: I was trying really hard not to say "should not assume" here, since it sounds like normative language. Hence, "can't assume" or something like that.

draft-ietf-snac-simple.xml
```diff
@@ -1081,34 +1080,38 @@
...
-        Note that DNS records for infrastructure-network-local hosts or services (e.g. records registered via SRP
-        or discovered by the SNAC router using mDNS) do not support DNSSEC protection.
+        is forwarded without TLS security to the upstream DNS resolver as configured by the infrastructure network.
+	    This enables an on-path eavesdropper in the infrastructure network to observe any DNS queries that the SNAC
+	    router forwards to the upstream DNS resolver, or an on-path attacker to spoof DNS responses to such queries.
+      </t><t>
+        Stub network hosts that implement DNSSEC <xref target="RFC9364"/> can perform DNSSEC validation of DNS responses
+        received from an upstream DNS resolver.
+	    Note that for locally-served zones {{RFC6303}} DNSSEC validation cannot work and is expected
+	    to be skipped by the SNAC router.
```
Abhayakara: This note no longer makes sense in this context. Previously we were suggesting that the SNAC router would do validation, but here we're talking about the stub network host. In principle we really don't need to talk about this: if the host does validation but doesn't understand locally-served zones, it may take the secure denial of existence to invalidate the response it gets from the local resolver, and there's not much we can do about this.

Abhayakara (changes requested):

draft-ietf-snac-simple.xml
EskoDijk: Whoops, SHOULD NOT say that. Will update.

draft-ietf-snac-simple.xml
EskoDijk: Right, I mixed up the two -- it should be the host/client. Since it's not for the SNAC router, I removed the note again -- this ought to be obvious for DNSSEC implementers for stub network hosts.

Abhayakara (approved):

PR #185, opened 2026-04-16
Provide a fix for the RFC 4861 problem of effectively fixed initial rtr interval.

### LINKED ISSUE #161: RFC 4861 "initial" (startup) router behavior timing is undesired in power outage reboot scenarios
This considers a home or shop power outage scenario, in which all of the SNAC routers may simultaneously reboot.

In this case, it's likely they follow the recommended "initial" behavior specified by RFC 4861, Section 6.2.4.

>  For the first few advertisements (up to
>   MAX_INITIAL_RTR_ADVERTISEMENTS) sent from an interface when it
>   becomes an advertising interface, if the randomly chosen interval is
>   greater than MAX_INITIAL_RTR_ADVERT_INTERVAL, the timer SHOULD be set
>   to MAX_INITIAL_RTR_ADVERT_INTERVAL instead.

This says to generate the random time in the interval [MinRtrAdvInterval, MaxRtrAdvInterval] which is [154, 206] sec in our case. Then if the random value is greater than `MAX_INITIAL_RTR_ADVERT_INTERVAL` , (16 sec), the interval is set to that max value instead, so 16 sec.

This means all of the N (e.g. 10) SNAC routers, if their boot time is approximately equal, will schedule their multicast at the same time t=16 sec.
Generally this is unwanted: we would prefer a random scheduling in some time period like [0,16] sec as startup behavior. Or something similar.

EskoDijk:
Note that the 2nd block is just whitespace corrections.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -422,7 +422,7 @@
 	    such prefixes in the RAs that it sends on the adjacent infrastructure link.
 	  </t>
       <t>
-		When a SNAC router enables its AIL interface as advertising interface and determines it needs to send unsolicited RAs
+		When a SNAC router enables its AIL interface as an advertising interface and determines it needs to send unsolicited RAs
 		per <xref target="RFC4861" section="6.2.4" sectionFormat="of"/>, it MUST use a random delay between
 		0 and MAX_INITIAL_RTR_ADVERT_INTERVAL, instead of the effectively fixed delay of MAX_INITIAL_RTR_ADVERT_INTERVAL
 		currently specified by <xref target="RFC4861"/>.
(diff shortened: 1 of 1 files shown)


EskoDijk:
The reason that RFC 4861 currently has an "effectively fixed" delay is the the random value is picked between 154 and 206 seconds, and then capped to  16 seconds max.  This results in always a 16-second delay.

By formulating generically in this section; it will apply in any cases also at SNAC router startup. So we don't have to say this anymore specifically in the startup/reboot section.

draft-ietf-snac-simple.xml
```diff
@@ -421,6 +421,12 @@
 	    to be valid on the stub network, even if those prefixes have been deprecated, and MUST include RIO options for all
 	    such prefixes in the RAs that it sends on the adjacent infrastructure link.
 	  </t>
+      <t>
+		When a SNAC router enables its AIL interface as advertising interface and determines it needs to send unsolicited RAs
```
```suggestion
		When a SNAC router enables its AIL interface as an advertising interface and determines it needs to send unsolicited RAs
```

EskoDijk: ok, done now!

PR #187, opened 2026-04-17
Editorial review comments by Fernando applied + grammar / fixes

### LINKED ISSUE #127: Editorial review comments by Fernando
This item collects all the editorial review comments by @fgont in one item. 

Could be addressed by 1 PR?

**** Editorial ****

* I would move the "Glossary" section right after Section 1.

* It would be great to have some kind of diagram in SEction 1 where you
sketch how this all works. e.g., I started reading the document assuming
that what you'd describe leads to a tree-like graph, where upstream
routers advertise to downstream routers. BUt as I continued reading
(still not done), it seems that SNAC routers advertise on both
interfaces. Also, can you have multiple layers of SNAC routers?
(it'd be great to get this high-level picture in the intro itself).

* Section 1.1:
> DNS-based DNS Service Discovery [RFC6763].

s/DNS-based DNS/

* Section 1.1:
> For stub network hosts that only consume cloud services this will not
> be an issue, but for stub networks that provide services,

"Cloud resources" usually implies "Internet resources". But in this case
you probably also mean resources in e.g. the infrastruccture network, or
some other stub network?

* Section 2:
> Stub Network:  A network link that is connected by one or more Stub 
> Routers to an AIL an infrastructure network,

Writeo?

* Section 5:
> Support for AILS

s/AILS/AILs/

* Section 5.1.1:
> Note: there can be layer two networks where neighbor discovery

s/layer two/layer-two/
s/neighbor discovery/Neighbor Discovery/

* Section 5.1.1:
> A prefix is considered to be advertised on the link if, when a
> Router Solicitation message ([RFC4861], Section 4.1) is sent, a
> Router Advertisement message is received in response which contains a
> prefix information option ([RFC4861], Section 4.6.2) for that
> prefix.
> 
> After an RA message containing a suitable prefix has been received, 
> it can be assumed for some period of time thereafter that that
> prefix is still valid on the link.  However, prefix lifetimes and
> router lifetimes are often quite long.  In addition to knowing that a
> prefix has been advertised on the link in the past, and is still
> valid, we must therefore ensure that at least one router that has
> advertised this prefix is still alive to respond to Router
> Advertisement messages.

Is this different from "good ol' slaac"?

* Section 5.2.1:
> Thread SNAC routers use bits from the Thread Extended PAN ID to 
> generate the ULA prefix's Global ID and Subnet ID.

Appropriate refs for Thread?

* Section 5.3:
> If the SNAC router is advertising a suitable prefix on any interface,
> any such prefixes MUST be advertised on that interface in the same
> Router Advertisement message that is advertising the suitable prefix,
> to avoid unnecessary multicast traffic.

It was hard for me to parse this para. Does it mean that if you are
advertising a PIO on one interface, that same prefix must be advertised
via RIOs in the same RAs that advertise PIOs?  Something else?

* Meta:
The document uses "Wi-Fi" and "WiFi" inconsistently.

* Section 5.5.1:
> WiFi tethering

I'd probably ad a reference for it.

* Section 6:
> stub networks

s/stub/Stub/

* Section6:
>    enables IPv6 hosts on the stub network to discover and reach with
>    IPv4 hosts on infrastructure,

s/with// ?

* Section 6:
>    This should be acceptable because hosts on the infrastructure network
>    should not be IPv4-only, since the SNAC router is providing IPv6
>    service on the infrastructure network that is suitable for
>    communicating using IPv6 to hosts on the SNAC network--there should
>    not be any hosts on the infrastructure network that can't communicate
>    with hosts on the stub network unless such hosts do not have an IPv6
>    stack at all.

Sentence seems to long, and hence hard to parse. Suggest breaking it 
down into smaller pieces.

* Section 6:
> and the return
>    packet will be destined for that IPv4 packet

s/for that IPv4 packet/for that IPv4 address/

* Section 6:
> SNAC routers that are not able to advertise an
>    infrastructure NAT64 service MUST NOT do so

I find this one hard to "parse" ("if they are not able, they must not do 
it"? -- well, they can't anyway?)

* Section 6.1:
> provides NAT64 service to IPv4-only hosts on the internet

s/internet/Internet/

* Section 6.2:
>    DNS resolver on the SNAC router MUST honor the 'CD' and 'DO' flag
>    bits if received in a request, since this indicates that the stub
>    resolver on the requestor intends to do DNSSEC validation.

Please add references for these. (RFC3225 and RFC4035?)

Also, please add references to NAT64 and DNS64 where appropriate.

* Section 8.  Services Provided by SNAC routers
I'd move this section way to the front of the document. The reader 
should be able to get the large picture without having to read through 
all the specific details.

* Appendix A.3.1:
>    In the case a host that is attempting to communicate with a device
>    using a site ULA prefix on a different link may choose a ULA address
>    as a source address. 

There seems to be a grammar error here?

* Appendix B:
>    *  In the options, with the exception of the options listed below,
>       SNAC routers MUST NOT send any RA options, since these other
>       options are for managing the network,

I'd probably rephrase a bit...

* Appendix C:
> xref target="I-D.ietf-6man-snac-router-
>       bit" TBD 

Fix?

EskoDijk:
also resolving some grammar issues and inserted missing 'tbd' content.
Close #127 

Note that 'Internet' is with capital I according to IETF guidelines.

Various spelling items like "Prefix Information option" and "Route Information Option" follow the naming as used by the source RFC.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -77,9 +77,9 @@
 	  process for stub routers, and can't assume that any particular stub router will always be available; rather, any stub
 	  router that is available must be able to adapt to current conditions to provide reachability.</li>
 	<li>
-	  Incompatible mechanisms: the medium of the stub network may not, for example, use IPv6 neighbor discovery to populate a
-	  neighbor cache. If the infrastructure network (as is typical) does use neighbor discovery, then bridging the two networks
-	  together would require some way of translating between neighbor discovery and whatever mechanism is used on the stub
+	  Incompatible mechanisms: the medium of the stub network may not, for example, use IPv6 Neighbor Discovery to populate a
+	  neighbor cache. If the infrastructure network (as is typical) does use Neighbor Discovery, then bridging the two networks
+	  together would require some way of translating between Neighbor Discovery and whatever mechanism is used on the stub
 	  network, and hence complicates rather than simplifying the problem of connecting the two networks.</li>
 	<li>
 	  Incompatible framing: if the stub network is a 6LoWPAN <xref target="RFC4944"/> network, packets on the stub network are
@@ -103,7 +103,7 @@
 	Stub networks may be globally reachable, or may be only locally reachable. A
 	host on a locally reachable stub network can only interoperate with hosts on the network link(s) to which it is connected.
 	A host on a globally reachable stub network should be able to interoperate with hosts on other network links in the same
-	infrastructure as well as hosts on the global internet.</t>
+	infrastructure as well as hosts on the global Internet.</t>
       <t>
 	It may be noted that just as one can plug several CE Router devices together in series to form multi-layer NATs, there is
 	nothing preventing the owner of a stub network router from attaching it to a stub network as if that network were its
@@ -142,13 +142,13 @@
 	  DNS <xref target="RFC6762"/>. As an example, when one host connected to
 	  a specific Wi-Fi network wishes to discover services on hosts connected to that same Wi-Fi network, it can do so using
 	  multicast DNS.  Similarly, when a host on some other network
-	  wishes to discover the same service, it must use DNS-based DNS Service Discovery <xref target="RFC6763"/>.  In both cases,
+	  wishes to discover the same service, it must use DNS-based Service Discovery <xref target="RFC6763"/>.  In both cases,
 	  "discoverable using DNS" means that the host has one or more entries in the DNS that serve to make it discoverable.</t>
 	<t>
 	  Discoverability is lumped in with reachability and addressability, both of which are essentially Layer 3 issues. The reason
 	  for this is that it does no good to automatically set up connectivity between stub network hosts and infrastructure
 	  hosts if the infrastructure hosts have no means to learn about the availability of services provided by stub network
-	  hosts. For stub network hosts that only consume cloud services this will not be an issue, but for stub networks that
+	  hosts. For stub network hosts that only consume Internet services this will not be an issue, but for stub networks that
 	  provide services, such as IoT devices on stub networks with incompatible media, discoverability is necessary in order for
 	  stub network connectivity to be useful.</t>
 	<t>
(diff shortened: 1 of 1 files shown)


darrendukes (approved):
Look good

jwhui (approved):

draft-ietf-snac-simple.xml
```diff
@@ -167,12 +167,21 @@
 	  An analysis of networks where SNAC routers could be attached is provided in <xref target="net-analysis"/>.</t>
       </section>
 
-      <section anchor="architecture"><name> Sample SNAC Router Aware Infrastructure Network</name>
+      <section anchor="architecture"><name>Sample SNAC Routers Deployment in a Home Network</name>
 
-      <t>A Stub network attached via SNAC routers the infrastructure network. The Infrastructure router with in the scope of this specification is an IPv6 (aware) router and provides different services in the network. These routers can be a CE, home gateway or a WiFI router. A SNAC router will connect as any another device in the infrastructure network, additionally providing addressability, discoverability and reachability functions for devices in the stub network.</t>
+      <t>A stub network is attached via one or more SNAC routers to an infrastructure network.
+		 An infrastructure router in the scope of this specification is an IPv6 (aware) router that provides different services in the network.
```
Abhayakara: "different services" probably means "various services." If so, that would be a better way to say this. This whole paragraph reads a bit strangely—really the infrastructure network is defined by the fact that it is infrastructure, not that it has an infrastructure router. The arrow points the other direction. Why do we even mention infrastructure routers here?

EskoDijk: Agree.  And I also found this  text confusing, but I didn't want to fully rewrite it in scope of this PR. (Just trying to clarify it a bit.)

Abhayakara: I've added an issue for adding a terminology section, so I don't think we want to fix this in this pull request.

draft-ietf-snac-simple.xml
```diff
@@ -167,12 +167,21 @@
...
+		 It can be a CE router, home gateway or Wi-Fi router. A SNAC router will connect as any another host in the infrastructure network,
+		 while additionally providing addressability, discoverability and reachability functions for hosts in the stub network.
+	     See <xref target="services-provided"/> for an overview of the specific services that a SNAC router implements to provide
+	     addressability, discoverability and reachability.</t>
 
-      <t>  This document describes mechanisms for connecting devices in Stub networks to the infrastructure using SNAC router as shown in <xref target="snac-rtr"/>. Although the use case described below is specific to home networks, the procedures apply for any type of infrastructure and stub network attachment scenarios.</t>
-      <t> A SNAC router receives and manages suitable IPV6 prefix on AIL preferably from the infrastructure router or from other SNAC routers in the network. In case addressing is not available, SNAC router provides and advertises it on the AIL.
+      <t>This document describes mechanisms for connecting IPv6 hosts in stub networks to the infrastructure newtork using
```
Abhayakara: s/newtork/network/

draft-ietf-snac-simple.xml
```diff
@@ -167,12 +167,21 @@
...
+		 SNAC routers as shown in <xref target="snac-rtr"/>. Although the use case depicted below is specific to home networks,
+		 the mechanisms apply for any type of infrastructure network and stub network attachment scenarios.</t>
+      <t>A SNAC router receives and manages a suitable IPv6 prefix for the AIL preferably from the infrastructure router
```
Abhayakara: I think this text is just wrong. It should say something more like:

A SNAC router looks for a suitable IPv6 prefix on the AIL. This prefix may be provided by the infrastructure router, or when no IPv6 infrastructure service is present, by some other SNAC router connected to the same AIL. If no suitable prefix is advertised on the AIL, the SNAC router will provide one in its own Router Advertisements.

The reason for this clarification is that we expect the SNAC router to always be sending RAs, so suggesting here that it only does so if there is no suitable prefix is incorrect. And it doesn't "receive and manage" a suitable prefix—it uses it if it's there, and if not it does the other thing. So I think the proposed text I've given here is more clear.

draft-ietf-snac-simple.xml
```diff
@@ -332,7 +341,7 @@
     <section anchor="support-ail">
       <name>Support for adjacent infrastructure links</name>
       <t>
-	This document assumes that the AIL supports neighbor discovery <xref target="RFC4861"/>, and specifically that
+	This document assumes that the AIL supports Neighbor Discovery <xref target="RFC4861"/>, and specifically that
 	routers and on-link prefixes can be advertised using Router Advertisement messages and discovered using Neighbor Solicitation messages. The stub
```
Abhayakara: routers are discovered using Router Solicitation messages, not neighbor solicitation messages.

Abhayakara (changes requested):
There are a couple of things to change before this is ready, but mostly it's ready.

draft-ietf-snac-simple.xml
EskoDijk: Now added the term 'infrastructure router' in the glossary - very short text, may be expanded later on.

draft-ietf-snac-simple.xml
EskoDijk: done now

draft-ietf-snac-simple.xml
EskoDijk: Updated to (more or less) the above proposal!

draft-ietf-snac-simple.xml
EskoDijk: Yes! fixed now.

EskoDijk:
> There are a couple of things to change before this is ready, but mostly it's ready.

Just fixed the items you flagged - let me know if it's complete now!  Then I'll merge it.

PR #190, opened 2026-04-20
Move 'maintenance across restarts' to a common section

### LINKED ISSUE #152: Section 5.2.1 discusses AIL prefix/addressability
Section 5.2.1, under 5.2, is supposed to consider only stub network prefix/addressability. However, it also talks about AIL aspects - see text below. This should be moved under 5.1 , right?  

Maybe the entire section 5.2.1 should move under 5.1 ?

> For example, suppose there are two SNAC routers connected to the same infrastructure link. When the first SNAC router is restarted, the second takes over providing an on-link prefix. Now the first router rejoins the link. It sees that the second SNAC router's prefix is advertised on the infrastructure link, and therefore does not advertise its own.
> 
> This behavior can cause problems because the first SNAC router no longer sees the on-link prefix it had been advertising on infrastructure as on-link. Consequently, if it receives a packet to forward to such an address, it will forward that packet directly to a default router, if one is present; otherwise, it will have no route to the destination, and will drop the packet.
> 
> To address this problem, SNAC routers SHOULD remember the last time a prefix was advertised across restarts. On restart, the router configures the prefix on its interface but does not advertise it in Router Advertisements. Devices that are still using that prefix will be seen as on-link to the router, and so packets will be delivered using ND on-link rather than forwarded to the default router.
> 
> When a SNAC router has only flash memory with limited write lifetime, it may be inappropriate to do a write to flash every time an RA beacon containing a prefix is sent. In this case, the router SHOULD record the set of prefixes that have been advertised on infrastructure and the maximum valid lifetime that was advertised. On restart, the router should assume that hosts on the infrastructure link have received advertisements for any such prefixes.
> 
> When possible, it is best if all SNAC routers serving a particular stub network use the same 64-bit prefix on the AIL. For example, Thread SNAC routers use bits from the Thread Extended PAN ID to generate the ULA prefix's Global ID and Subnet ID. The Global ID generation conforms to [RFC4193] because the Extended PAN ID is generated randomly using the same mechanism that is specified in RFC 4193 for the ULA prefix bits.
>

### LINKED ISSUE #155: Maintenance across SNAC router restarts
The scenario that this section addresses is somewhat specific to the Thread border router use case. Ironically, we decided that this doesn't work very well and wound up using a different approach. I think this text is left over from before we did that. Given that this approach didn't work for Thread, it seems inappropriate to advise this behavior for other use cases that we don't yet know about. So I think we should remove this section.

Abhayakara:
This pull request moves the 'maintenance across restarts' section to the beginning and moves some common paragraphs from the AIL support section (previously section 5) before that to serve as an introduction, since those paragraphs were also generally applicable.

It further rewrites this section to avoid specifying behavior that didn't work well and is no longer used in existing deployments.

Closes #152
Closes #155

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -229,8 +229,10 @@
       </section>
     </section>
 
-    <section>
+    <section anchor="glossary">
       <name>Glossary</name>
+	<t>This section contains a glossary of terms used in this document.
+	   See <xref target="conventions"/> for document conventions and remaining terms and definitions.</t>
       <dl>
 	<dt>Node:</dt>
 	<dd>A device that implements IPv6.</dd>
@@ -344,7 +346,7 @@
       </dl>
     </section>
 
-    <section>
+    <section anchor="conventions">
       <name>Conventions and Terminology Used in This Document</name>
       <t>
 	The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
@@ -355,6 +357,9 @@
 	The terms "resolver" and "stub resolver" used in this document are defined in <xref target="RFC9499"/>.
 	"DNS resolver" is used as a synonym for "resolver" to clearly point out the DNS context of this term.
       </t>
+      <t>
+	The term "advertising interface" is used as defined in <xref target="RFC4861" section="6.2.2"/>.
+	  </t>
     </section>
 
     <section anchor="common-maint">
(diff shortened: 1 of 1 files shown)


EskoDijk:
@Abhayakara I can provide some more corrections on the text. 
For the term "Thread router" we probably mean a SNAC router that acts as Thread Border Router?  (Which is itself a 6LoWPAN Border Router, more generally -- this term is better estabished in IETF.)

Abhayakara:
That would be great—thanks!

Abhayakara:
Er, @EskoDijk in case the previous doesn't show up.

EskoDijk:
@Abhayakara I added a few proposed corrections in the 2 additional commits. But, seeing a problem in the new text: the Section 5 intro text points out things that were supposed to be applicable to the (now) Section 6:

* use of term "advertising interface"
* When a SNAC router sends a Router Advertisement (RA) message as
   specified in any of the below subsections, it MUST NOT use multiple
   RA messages with each containing a subset of the options.  
    * this was intended for Section 6.x subsections. But now it's in section 5.

Not sure how this was intended to be placed?

Abhayakara:
The sections it applies to are still below.

EskoDijk:
> The sections it applies to are still below.

Sure, but quite misleading for someone with my brain. I've now generalized it - "advertising interface" becomes just another term/definition. And the rule applies to the whole document.

Now reviewing one more time the complete change.

EskoDijk:
Looks ok now; I'll merge it!

PR #191, opened 2026-04-20
internal links from constants to definitions

### LINKED ISSUE #137: Manifest constants in document should have reference links so that we don't have to scroll around
We have a number of constants defined in the document which are mentioned only as constants in the text where the implementor is urged to apply them, which means that when implementing, the reader has to scroll through the document. So let's add reference links to all of these constants to address that issue.

Abhayakara:
Closes #137


EskoDijk (approved):
That looks nice! I'll merge it now.

PR #200, opened 2026-04-30
Editorial review as part of WGLC, by Esko.

EskoDijk:
Provides a complete editorial review with corrections and suggestions.

Notably, any normative requirement that was posed on multiple SNAC routers at once is now converted to a requirement on a single SNAC router.

Some terms were updated to match already-defined terms, notably "OSNR prefix" is more heavily used now to be clear what stub network prefixes are actually advertised externally.


jwhui (approved):

EskoDijk:
(just rebased the PR.)

PR #207, opened 2026-05-01
Clarified requirements on default router advertisement on stub network; added requirements for withdrawing default route again.

### LINKED ISSUE #192: Requirements on withdrawing default route on stub network
Current text -09 has:

   The SNAC router MAY advertise itself as a default router on the stub
   network, if it itself detects that a default route is present on the
   AIL.  

There is no mention of behavior for the case that the SNAC router does this and the default route is withdrawn on the AIL. If this is done, there needs to be perhaps a requirement on how to withdraw the default route also -> e.g. immediately set the Router Lifetime field of the RA to 0 - in case RAs are used on the stub network? Or does it "mirror" the Router Lifetime of the AIL ?

EskoDijk:


draft-ietf-snac-simple.xml
```diff
@@ -831,9 +831,18 @@
...
 	<t>
-	  The SNAC router MAY advertise itself as a default router on the stub network, if it itself detects that a default route is
-	  present on the AIL. In some cases it may not be desirable to advertise reachability to the Internet as a whole; in this case the SNAC
-	  router is not required to advertise itself as a default router.
+	  The SNAC router SHOULD advertise itself as a default router on the stub network, if it detects that a default route is
+	  present on the AIL. In some cases it may not be desirable to advertise reachability to the Internet as a whole; for such cases the SNAC
+	  router MAY be administratively configured to not advertise itself as a default router, or it MAY use an automated
+	  policy to make such a decision.
+	</t><t>
+	  A SNAC router MUST NOT advertise a longer default route lifetime on its stub network than the default router lifetime
```
jwhui: The SNAC router will typically send messages advertising a default route on the stub network at different times than the RAs received on the infrastructure network.

Is the requirement here that the SNAC router must maintain take into account the time since the last RA was received when advertising the default route lifetime on the stub network?

EskoDijk: That's right - I've now rephrased this to:

	  A SNAC router MUST NOT advertise a default route lifetime on its stub network that is higher than the remaining default router lifetime on the AIL.

("Remaining" should imply that the lifetime is counted down since the last reception of an RA.)

PR #208, opened 2026-05-01
Add missing state transition for STATE-DEPRECATED to STATE-SUITABLE.

EskoDijk:
Also few caps fixes for RA.


EskoDijk:
@Abhayakara An issue from review: the state transition from 'deprecated' back to 'suitable' was missing.  I think this was implicitly already there, but more clear to make that explicit since we also describe another transition out of the deprecated state.

PR #209, opened 2026-06-23
AD review: ### DHCP or DHCPv6 ? 

EskoDijk:
-> uniform term use and include early refs to RFC 9915.


evyncke (approved):

draft-ietf-snac-simple.xml
```diff
@@ -1195,7 +1195,7 @@
 	    router forwards to the upstream DNS resolver, or an on-path attacker to spoof DNS responses to such queries.
       </t><t>
         Stub network hosts that implement DNSSEC <xref target="RFC9364"/> can perform DNSSEC validation of DNS responses
-        received from the DNS recursive resolver on the SNAC router.
+        received from the DNS resolver on the SNAC router.
```
Abhayakara: Please see my comment on the mailing list.

Abhayakara (changes requested):
Aside from the recursive resolver question this all looks good. I would suggest taking that out of this pull request and waiting for Eric to respond before doing a new pull request just on that item.

draft-ietf-snac-simple.xml
EskoDijk: I've now reverted this change - 2nd commit.

jwhui (approved):

Abhayakara (approved):

PR #212, opened 2026-06-29
Change RFC 7084 reference to 7084bis draft. 

EskoDijk:
As flagged in Tim's shepherd review: this avoids a downref and is more up-to-date, however adds a normative ref to a draft. For WG consideration.


evyncke (approved):
I support the change 1) pointing to recent RFC (when published) 2) avoid a downref (though never critical).

This I-D will anyway be delayed for the normative references to the DNSSD drafts

jwhui (approved):

PR #214, opened 2026-07-17
AD review: Sections 3, 4 and 5. Also introduces a new section header structure.

EskoDijk:
New section header structure is to more clearly indicate the topics discussed and avoid the wording 'maintenance' where regular operation was intended.

### DIFF
--- a/draft-ietf-snac-simple.xml
+++ b/draft-ietf-snac-simple.xml
@@ -368,28 +368,27 @@
       <dl>
 	<dt anchor="STALE_RA_TIME">STALE_RA_TIME (default: 10 minutes):</dt>
 	<dd>
-	  The amount of time that can pass after the last time a Router Advertisement (RA) message from a particular router has
+	  The amount of time that can pass after the last time a Router Advertisement (RA) message from a particular IPv6 router has
 	  been received before it is assumed the router is no longer present. This is a stopgap in case the router is reachable but has
-	  silently stopped advertising a prefix; this situation is unlikely, but if it does happen, new devices joining the
+	  silently stopped sending RA messages; this situation is unlikely, but if it does happen, new devices joining the
 	  infrastructure network will not be able to reach devices on the stub network until the SNAC router decides that the router
-	  that advertised the suitable prefix is stale.</dd>
+	  that previously advertised the suitable on-link prefix on the AIL is stale.</dd>
 	<dt anchor="STUB_PROVIDED_PREFIX_LIFETIME">STUB_PROVIDED_PREFIX_LIFETIME (default: 30 minutes):</dt>
 	<dd>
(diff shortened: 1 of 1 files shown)


draft-ietf-snac-simple.xml
```diff
@@ -430,21 +429,28 @@
...
+        This section specifies requirements that are common to both AIL and stub network links, as well as general
+        requirements for SNAC routers.</t>
+    <section>
+      <name>Support for IPv6 Neighbor Discovery</name>
       <t>
-	This document assumes that the AIL supports Neighbor Discovery <xref target="RFC4861"/>, and specifically that
-	routers and on-link prefixes can be advertised using Router Advertisement messages and discovered using Router Solicitation messages. The stub
-	network link may also support this, or may use some different mechanism. This section specifies how advertisement of the
-	on-link prefix for such links is managed.</t>
+	This specification requires that the AIL can support IPv6 Neighbor Discovery (ND) <xref target="RFC4861"/>, and specifically that
```
```suggestion
	This specification requires that the AIL supports IPv6 Neighbor Discovery (ND) <xref target="RFC4861"/>, and specifically that
```

draft-ietf-snac-simple.xml
```diff
@@ -474,35 +483,48 @@
...
+	  Similarly, it can be the case that while a SNAC router is advertising the sole OSNR prefix on the stub network,
+      it is rebooted or otherwise offline for a period of time.
+      In this case, if other SNAC routers are present, they can continue to advertise routes to that prefix on the AIL.
+	  Ideally, the original OSNR prefix would be preserved by the other SNAC router(s) until the SNAC router that originally
+      advertised it returns. However, any coordination method to do so is out of scope of this specification.
+      In absence of such a coordination method, if a host on the stub network multicasts a Router Solicitation message
+      to the stub network link and the SNAC router (that was previously advertising the OSNR prefix) is still not present,
+      or if the OSNR prefix is in danger of expiring, another SNAC router will eventually take over and
+	  advertise its own OSNR prefix. In this situation, all SNAC routers that remember the old OSNR prefix continue
+      advertising a route to it on the AIL until the OSNR prefix expires.
```
```suggestion
      advertising a route to it on the AIL until the old OSNR prefix expires.
```

evyncke (commented):
Reviewing a XML PR is really difficult. Please merge all PRs, convert to TXT and use the tool to make a side-by-side diff and send to me (copying snac@ietf.org).

draft-ietf-snac-simple.xml
EskoDijk: @jwhui This is intended as "can support" and not "support" (the latter suggests it's already supported I think?). So the link needs to be capable of supporting the things that IPv6 ND does to the link, like multicast. It's not required that the link already has a working IPv6 ND protocol node on it.

"is able to support" is an alternative phrasing, maybe I'll just use that.

PR #215, opened 2026-07-17
AD review: Section 6 (comments partially addressed).

EskoDijk:


Abhayakara:
These differences are really hard to review because of the way that github presents diffs. We have better tools in author-tools.ietf.org. It might make more sense to simply produce a new version of the document with these changes and let us review it that way.

draft-ietf-snac-simple.xml
```diff
@@ -446,16 +446,22 @@
...
 	  one or more other SNAC routers or by other router types, it is advised that the SNAC router takes appropriate actions such as warning
 	  the user or ceasing to operate as a SNAC router until the situation is resolved.
 	  How a stub router might detect this situation is however out of scope for this document.
 	  </t>
+      <t>
+      Similarly, a SNAC router that is simultaneously active on more than one AIL is explicitly out of scope of this
+      document. This means that a SNAC router also never provides routing between AILs when connected to more than one such link.</t>
+      <t>
+      A SNAC router MAY during its operation select a different AIL interface to operate on.
+      For example, switching from Wi-Fi to Ethernet when its Ethernet AIL interface gets connected.</t>
```
jwhui, line 470: Should we explicitly state behavior for the old (non-advertising) interface? In some cases, this may not be possible if it was a hard switch over, but in the cases where it is possible?

draft-ietf-snac-simple.xml
```diff
@@ -883,17 +905,15 @@
...
-	  combine the OSNR route advertisements (RIOs) and the on-link prefix advertisement (PIO) in the same Router
-	  Advertisement message, to avoid unnecessary multicast traffic.
+	  A SNAC router MUST advertise reachability to stub network OSNR prefixes on its AIL interface using Route Information Options (RIOs) in Router
+	  Advertisement (RA) messages. As specified in <xref target="common-maint"/>, these options are combined with other RA options, to avoid unnecessary multicast traffic.
 	</t><t>
 	  Each stub network will have some set of prefixes that are advertised as on-link for that network. A SNAC router connected
-	  to that stub network SHOULD advertise reachability to all such prefixes on the AIL to which it is attached using Router
-	  Advertisements.
+	  to that stub network SHOULD advertise reachability to all such prefixes on the AIL to which it is attached.
+      The exception case for this requirement is where the number of prefixes becomes too high to fit into a Router Advertisement message: then, a SNAC router can leave out deprecated prefixes and/or prefixes that become invalid soonest.
```
jwhui: Should we use MAY here?
```suggestion
      "The exception case for this requirement is when the number of prefixes exceeds what can fit into a single Router Advertisement message; in this case, the SNAC router MAY omit deprecated prefixes and/or prefixes with the shortest remaining valid lifetimes."
```

draft-ietf-snac-simple.xml
```diff
@@ -858,8 +873,15 @@
 	    </ul>
 	    <t>
 	      The SNAC router MUST NOT use a prefix once the DHCPv6-PD client has determined that it is no longer valid. If the
-	      DHCPv6-PD client provides a new prefix, and the old prefix is still valid, the SNAC router SHOULD explicitly deprecate
-	      the old prefix at the same time that it first advertises the new prefix.</t>
+          DHCPv6-PD client provides a new prefix, and the old prefix is still valid, then:</t>
+          <ul>
+          <li>in a stub network where a single OSNR prefix constraint applies, the SNAC router MUST deprecate
+              the old OSNR prefix at the same time that it first advertises the new OSNR prefix.</li>
+          <li>in a stub network where a multiple OSNR prefixes constraint applies, the SNAC router MAY deprecate
```
```suggestion
          <li>in a stub network where multiple OSNR prefix constraints apply, the SNAC router MAY deprecate
```

draft-ietf-snac-simple.xml
```diff
@@ -446,16 +446,22 @@
 	  If neighbor discovery, including Router Advertisements, is being used on the stub network, the SNAC router MUST similarly join these two multicast groups on the stub network interface as well.
 	</t>
       <t>
-      A network topology of a single stub network connected via SNAC routers to more than one AIL is explicitly out of scope of this document;
+      A network topology of a single stub network connected via multiple SNAC routers to more than one AIL is explicitly out of scope of this document;
 	  such a topology will likely have a negative effect on the reachability and discoverability of hosts.
       If a SNAC router is able to determine that an AIL (other than its own) is already in use for its stub network, for example provided by
 	  one or more other SNAC routers or by other router types, it is advised that the SNAC router takes appropriate actions such as warning
 	  the user or ceasing to operate as a SNAC router until the situation is resolved.
 	  How a stub router might detect this situation is however out of scope for this document.
```
```suggestion
	  How a SNAC router might detect this situation is however out of scope for this document.
```

evyncke (commented):
Reviewing a XML PR is really difficult. Please merge all PRs, convert to TXT and use the tool to make a side-by-side diff and send to me (copying snac@ietf.org).

EskoDijk:
> Reviewing a XML PR is really difficult. Please merge all PRs, convert to TXT and use the tool to make a side-by-side diff and send to me (copying snac@ietf.org).

Yes, that's the plan - will do that hopefully soon. With a few Pr updates based on review input.

draft-ietf-snac-simple.xml
EskoDijk: @jwhui Good question - created a new issue for this: https://github.com/ietf-wg-snac/draft-ietf-snac-simple/issues/217  - to see what to do with that post-WGLC.

draft-ietf-snac-simple.xml
EskoDijk: @jwhui  I've now replaced by MAY - though practically it amounts to the same thing. Having normative language in an exception case of a SHOULD might also be scrutinized by IETF reviewers; but I'm not aware of any policies on that.

draft-ietf-snac-simple.xml
EskoDijk: @jwhui I've now changed to below text using quotes, because it refers back to one of two named constraints higher up in the text that use this specific wording.

> in a stub network where the "single OSNR prefix constraint" applies, the SNAC router MUST deprecate
> the old OSNR prefix at the same time that it first advertises the new OSNR prefix.
> in a stub network where the "multiple OSNR prefix constraint" applies, the SNAC router MAY deprecate

It could also be called a "case" instead of a "constraint".

PR #216, opened 2026-07-18
AD review updates: Section 7 and up.

EskoDijk:


draft-ietf-snac-simple.xml
```diff
@@ -1041,22 +1041,22 @@
 	on infrastructure to communicate to IPv6 hosts on the stub network.
       </t>
       <t>
-	This should be acceptable, because hosts on the infrastructure network that need to access stub network hosts should
+	This should be acceptable, because hosts on the infrastructure network that need to access stub network IPv6 hosts should
```
Abhayakara: Presumably this is to differentiate from those pesky IPv4 or chaosnet stub network hosts? I would prefer not to say this—I don't think it's necessary and winds up begging a question. If this really needs clarification I would quibble more with "should not be IPv4-only" and change the sentence to:

...hosts on the infrastructure network that need to **communicate with** hosts on the stub network **should be capable of using IPv6** to do so, and of course it is necessary that hosts on the stub network be capable of communicating using IPv6 since SNAC routers do not provide IPv4 service on their stub networks.

draft-ietf-snac-simple.xml
```diff
@@ -1172,20 +1177,22 @@
...
+	  and is tied to the SNAC router's IPv4 address. Therefore, each NAT64 egress is not equivalent.
 	</t><t>
-	  This specification requires the stub network host itself to perform DNS64 <xref target="RFC6147"/> synthesis, as
-	  needed. A SNAC router does not provide DNS64 synthesis. Instead, it MUST provide an ipv4only.arpa answer that
-	  advertises the NAT64 prefix for that SNAC router, and MUST provide an explicit route to that NAT64 prefix on the stub
-	  network using RA or an equivalent protocol for that stub network type.
+	  This specification requires the stub network host itself to perform local IPv6 address synthesis, as needed,
+	  analogous to the customer-side translator (CLAT) function of 464XLAT <xref target="RFC6877"/>. This is the
+	  host-side equivalent of the network-based DNS64 <xref target="RFC6147"/> function. A SNAC router does not
+	  provide DNS64. Instead, it MUST provide an ipv4only.arpa answer <xref target="RFC7050"/> <xref target="RFC8880"/>
```
Abhayakara: Hm, shouldn't we be using PREF64 here and not ipv4only.arpa?

draft-ietf-snac-simple.xml
```diff
@@ -1172,20 +1177,22 @@
...
+	  that advertises the NAT64 prefix for that SNAC router, and MUST provide an explicit route to that NAT64 prefix
+	  on the stub network using RA or an equivalent protocol for that stub network type.
 	</t><t>
 	  In constrained networks it can be very useful if stub network DNS resolvers provide the information required to do DNS64
 	  translation in the answer to the AAAA query. If the answer to an AAAA query comes back with "no data" (not NXDOMAIN), this
 	  suggests that there may be an A record. In this case, the DNS resolver in the SNAC router SHOULD attempt to look up an A record on
 	  the same name. If such a record exists, the DNS resolver SHOULD return no data in the Answer section of the DNS response, and
 	  SHOULD provide any CNAME records that were involved in returning the "no data" answer to the AAAA query, and SHOULD
-	  provide any A records that were ultimately returned, in the Additional section. The response message SHOULD also include an
+	  provide any A records that were ultimately returned, both in the Additional section. The response message SHOULD also include an
```
Abhayakara: both is incorrect here because it indicates that there are only two records, even though both antecedents are themselves plural. Also, the CNAME records used to find the AAAA answer go in the answer section, not the additional records section. So the document with this change is simply incorrect.

If we need to make a change here, I would rewrite it as folllows:

If the answer to the AAAA query returns no AAAA records, any CNAME records returned in the answer section for that query SHOULD be returned in the answer section. If an A query was done, and produced an answer, any CNAME records and A records returned in the A query response should be returned in the Additional section of the SNAC resolver's AAAA query response.

draft-ietf-snac-simple.xml
```diff
@@ -1364,7 +1383,7 @@
 	<name>Use on a managed network</name>
 	<t>
 	  In this scenario, a non-expert user attaches a SNAC router to an infrastructure network that's managed and provides both IPv6 and IPv4.
-	  This network has correctly deployed RA Guard and/or port-based access control.
+	  This network has correctly deployed RA guard and/or port-based access control.
```
Abhayakara: This is actually written "RA-Guard" in RFC7113.

draft-ietf-snac-simple.xml
```diff
@@ -1379,15 +1398,20 @@
 	  will be correctly routed.
 	</t>
 	<t>
-	  It's worth noting that IPv4 devices that act similarly to SNAC routers, using NAT64, already exist and may indeed use the stub network
-	  functionality to support internal connections that aren't even apparent to the user. In this case the SNAC router is not relying on
-	  RA to function because it's using its IPv4 address and NAT64 to provide connectivity, so there is no management issue even if RA is
-	  blocked. This is a reasonable use case for IPv6, and this specification does in fact enable this use case.
+      It's worth noting that devices which behave like SNAC routers already exist: they attach to the infrastructure network as
```
Abhayakara: "that" is correct here, not "which." Which is used to point to specific instances; that is used to point to a general set.

Abhayakara (changes requested):

draft-ietf-snac-simple.xml
EskoDijk: Ok, the majority of terms was "RA guard" in our draft. But happy to change all these entries to official!

draft-ietf-snac-simple.xml
Abhayakara: Yeah, sorry about that, that's my fault. I don't think I've looked at the RA Guard document since 2012-ish. But I think we should indeed use the official term.

draft-ietf-snac-simple.xml
EskoDijk: The original text says there is no data in the Answer section, so I think everything goes into the Additional section then?   "Both" was to clarify that both the CNAME records mentioned just before and the A records mentioned here go into the Additional section.  But "both" is indeed the wrong word for this ... 
Still, a change would be needed to clarify (just-in-case) that the CNAME things (if any) go into Additional.

draft-ietf-snac-simple.xml
EskoDijk: This is on the stub network side - we required PREF64 to look for NAT64 prefixes on the AIL side.
On the stub, if it's a link that supports PREF64, we use that; otherwise the stub-net-specific mechanism. We have existing text for that:

```
Some network media may provide their own mechanisms for advertising NAT64 service to the stub network. If such a mechanism
	is available, a SNAC router MUST use the mechanism provided by the network medium used on the stub network to advertise
	NAT64 service. Otherwise, NAT64 service MUST be advertised on the stub network using the PREF64 Router Advertisement option
	<xref target="RFC8781"/>.
```

So to cover any and all cases, if the resolver provides the ipv4only.arpa answer, then it covers any cases where PREF64 is not used. (E.g. Thread?)
If we rather want to delegate to the "specific mechanism" for the stub network instead, we can change this text along those lines in agreement with the above quote.

draft-ietf-snac-simple.xml
EskoDijk: we can remove IPv6 here, ok. It was just to emphasize that these hosts are necessarily IPv6.
The scope of the draft is now more clearly stated (by these combined PRs) to be IPv6 for the stub network only. (No IPv4.)

But we already say the same in the paragraph above this one so happy to just remove it here.

draft-ietf-snac-simple.xml
Abhayakara: Right, the original text is also wrong! :)

In fact, a correct DNS response in this context would include any CNAMEs returned in the A query to the infrastructure resolver in the answer section of the response sent by the SNAC resolver. This is what RFC1035 says.

The A query is an additional query, so any responses we get in the answer section of the response to _that_ query, whether CNAME or A, go in the additional section. But actually the CNAMEs are only required if we eventually got an A record; otherwise there's no point in sending them because the client can't rely on them as an indication that there is no A answer.

We should also leave out any CNAME records from the A query response that recapitulate CNAME records from the AAAA query response, since they are redundant. In most cases, this will mean no CNAME records in the additional section. We could get this detailed here, and maybe we should, although this is .. recapitulating RFC1035, section 6.2. Which actually refers back to RFC1034. So. I think it's worth giving the top-level points here, but to some extent we have to rely implementors to refer to RFC1034 and RFC1035.

draft-ietf-snac-simple.xml
Abhayakara: Err.. Actually I kind of want to change the previous paragraph as well:

Although NAT64 <xref target="RFC6146"/> provides IPv6-only hosts with a way to reach IPv4-only hosts, there is no easy way for an IPv4-only host to use NAT64 to originate communication with a host using IPv6. Therefore, a SNAC router enables IPv6 hosts on the stub network to discover and reach IPv4 hosts on infrastructure, but does not provide a way for IPv4-only hosts on infrastructure to communicate to IPv6 hosts on the stub network.

This should be acceptable, because hosts on the infrastructure network that need to communicate with hosts on the stub network are expected to be capable of using IPv6 to do so, and of course it is necessary that hosts on the stub network be capable of communicating using IPv6 since SNAC routers do not provide IPv4 service on their stub networks.

draft-ietf-snac-simple.xml
```diff
@@ -1110,35 +1110,40 @@
 	out.</t>
       <t>
 	A further complication is that in some cases, some SNAC routers connected to the stub network may not be able to
-	advertise an infrastructure-provided NAT64 prefix, while others may. In this case, when an infrastructure-provided NAT64
+	advertise an infrastructure-provided NAT64 prefix, while others may.
+    This can happen for example due to a stub network technology limitation on the number of (NAT64) prefixes that can
```
```suggestion
    This can happen, for example, due to a stub network technology limitation on the number of (NAT64) prefixes that can
```

draft-ietf-snac-simple.xml
```diff
@@ -1110,35 +1110,40 @@
...
     on the stub network MUST stop attempting to advertise NAT64 service itself until the moment that there is no more NAT64
     service advertised on the stub network.</t>
       <t>
 	For stub network technologies that support the advertising of a NAT64 service with an associated preference level,
 	the below rules for preference level selection MUST be used.
-	For stub network technologies that don't support this (for example, technologies using the PREF64 option), these rules do
-	not apply.</t>
+    Although the details of such a stub network technology's service advertisements are out of scope of this document,
+    it is assumed that at least the three levels of 'low', 'medium', and 'high' can be distinguished with it.
+	For stub network technologies that don't support NAT64 preference level (for example, technologies using the PREF64 option),
```
```suggestion
	For stub network technologies that don't support NAT64 preference levels (for example, technologies using the PREF64 option),
```

